PR #1, opened 2019-03-30
add new projects to cli documentation

halfhorst:
updating cli docstring (and thus cli help docs) to include new projects

### DIFF
--- a/src/vivarium_cluster_tools/cli.py
+++ b/src/vivarium_cluster_tools/cli.py
@@ -21,7 +21,9 @@ def psimulate():
 @click.option('--result-directory', '-o', default=None,
               help='The directory to write results to. A folder will be created in this directory with the same name '
                    'as the configuration file.')
-@click.option('--project', '-P', type=click.Choice(['proj_cost_effect', 'proj_csu', 'proj_cost_effect_diarrhea']), default='proj_cost_effect',
+@click.option('--project', '-P', type=click.Choice(['proj_cost_effect', 'proj_cost_effect_diarrhea',
+                                                    'proj_cost_effect_dcpn', 'proj_cost_effect_conic', 'proj_csu']),
+              default='proj_cost_effect',
               help='The cluster project under which to run the simulation.')
 @click.option('--peak-memory', '-m', type=int, default=3,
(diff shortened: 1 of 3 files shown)


src/vivarium_cluster_tools/cli.py
```diff
@@ -21,7 +21,9 @@ def psimulate():
 @click.option('--result-directory', '-o', default=None,
               help='The directory to write results to. A folder will be created in this directory with the same name '
                    'as the configuration file.')
-@click.option('--project', '-P', type=click.Choice(['proj_cost_effect', 'proj_csu', 'proj_cost_effect_diarrhea']), default='proj_cost_effect',
+@click.option('--project', '-P', type=click.Choice(['proj_cost_effect', 'proj_cost_effect_diarrhea',
```
kfwilson: now that I see this in a PR instead of me just hacking an extra project in to make diarrhea work, maybe we should have a global list for the projects? Just so that run and restart don't end up with a mismatched list?

halfhorst: I don't have an issue with that, it fits the bill of a global variable

collijk: Pull it out into a globals.py file so we can use it in the runner as well.

src/vivarium_cluster_tools/runner.py
```diff
@@ -44,9 +45,7 @@ def uge_specification(peak_memory, project, job_name):
         if os.environ['SGE_CLUSTER_NAME'] == "dev":
             project = None
         else:  # prod or new cluster
-            assert project in ['proj_cost_effect', 'proj_cost_effect_diarrhea',
-                               'proj_cost_effect_dcpn', 'proj_cost_effect_conic',
-                               'proj_csu'], 'Script only for use by sanofi and cost effectiveness teams.'
+            assert project in CLUSTER_PROJECTS, 'Script only for use by sanofi and cost effectiveness teams.'
```
kfwilson: Can we tell the user what valid projects there are? This wouldn't be a super helpful error message

halfhorst:
> looks good. the only thing would be it might be nice to tell the user what the valid cluster projects are if they give an invalid one

done

PR #2, opened 2019-03-30
Feature/distributed runner docs

halfhorst:
setup basic readthedocs integration and add documentation for installing this package and using it to run simulations.

you can see the docs: https://readthedocs.org/projects/vivarium-cluster-tools/

make sure you are looking at the version corresponding to this branch . latest doesn't work and stable either doesn't work or shows the readme


docs/source/distributed_runner.rst
```diff
@@ -0,0 +1,45 @@
...
+
+    psimulate run /path/to/your/model/specification  /path/to/your/branch
+
+By default, output will be saved in ``/share/scratch/users/{$USER}/vivarium_results``. If you want to save the
+results somewhere else you can specify your output directory as an optional argument
+
+.. code-block:: console
+
+    psimulate run /path/to/your/model/specification /path/to/your/branch -o /path/to/output
+
```
kfwilson: Oh one thought: maybe add a sentence saying the runner will create a folder named the time your run starts in that output directory? I had a few questions about whether if you give two distributed runs the same output path, will they overwrite outputs

halfhorst: that's a good point

PR #3, opened 2019-04-01
Feature/basic yaml

halfhorst:
This section details enough YAML syntax to read our configs.

There are references to two other sections in there that I'll fill in when those get PR'd


docs/source/yaml_basics.rst
```diff
@@ -0,0 +1,122 @@
...
+
+Comments
+--------
+
+YAML comments are denoted with the pound symbol ``#``, and can be placed anywhere, but must be separated from the
+preceding token by a space. For example, adding a comment to the configuration from above looks like this:
+
+.. code-block:: yaml
+
+    components:
```
collijk: Still strongly advise against using the components block for examples.  What matters for users is the ``configuration`` block, as that passes into individual components via the builder and gets modified by the branches file.  The ``components`` section has meaningful semantics that are not yaml specific.

docs/source/yaml_basics.rst
```diff
@@ -0,0 +1,122 @@
...
+                - BasePopulation()
+                - Mortality()
+                - FertilityCrudeBirthRate()
+
+This will be interpreted as
+
+.. code-block:: python
+
+    {components: {
+            vivarium_public_health: {
```
collijk: e.g. this is not at all how the framework interprets this block.  it gets interpreted as 
```
[vivarium_public_health.population.BasePopulation(), vivarium_public_health.population.Mortality(), vivarium_public_health.population.FertilityCrudeBirthRate()]
```

docs/source/yaml_basics.rst
halfhorst: I swapped this out , but not the list example because the only place I see those is when listing components. Configs are mappings as far as I can see

docs/source/yaml_basics.rst
```diff
@@ -0,0 +1,127 @@
...
+            population_size : 1000,
+            age_start: 0,
+            age_end: 30
+            }
+        }
+    }
+
+Lists
+-----
+
```
kfwilson: this section is a nice addition rather than just bringing up lists in the branches doc with no prior reference like before

PR #4, opened 2019-04-02
Feature/branches doc

halfhorst:
This section is on branches. Need to link to it from the YAML basics section


halfhorst:
On hold

docs/source/branch.rst
```diff
@@ -0,0 +1,52 @@
+
+The Branches File
+=================
+
+When trying to answer the research questions that drove the construction of a model and a configuration, it is often
```
collijk: Both want to vary parameters, but also capture uncertainty.  Capturing uncertainty should be the first topic, since it's simpler (doesn't involve counterfactuals, just repeated runs of the same sim).

Outline of what should be here:

Section on uncertainty:
Subsection on varying parameter uncertainty with an example branches file with just the input draw count
Subsection on varying stochastic uncertainty with an example branches file with just the random seed count
- Note about how varying stochastic uncertainty allows aggregations, effectively letting us increase the sample size for a given input draw. (e.g. 1 sim with 1M people == 100 sims with 10k people).
Subsection on varying both at the same time with description on how we arrive at the number of runs and example model spec w/ both input_draw_number and random_seed_count  args.

Section on parameter variations (or scenarios):
Intro: "the branches section defines a set of scenarios.  We'll describe a number of ways you might want to construct simple or complex scenario specifications..."
Subsection with single parameter variation and example model spec with just the branches key.  
Subsection about how this combines with the input_draw_number/random_seed_count.
Subsection with variation of two parameters.  Explanation of how this turns into multiple scenarios.

collijk: Final section on complex configs (use shigella example)

halfhorst:
there may be some spots where links would be useful. Also, I was thinking a note at the top-level would be good that gives an outline of the recommended order to read sections in, or a short description or something

docs/source/branch.rst
```diff
@@ -0,0 +1,218 @@
...
+allowing us to vary the input draws from the Global Burden of Disease (GBD) used as well as the random seeds used by the
+simulation.
+
+Data Uncertainty
+^^^^^^^^^^^^^^^^
+Our simulations primarily rely on GBD results which are produced in draws to estimate uncertainty. In order to reflect
+the GBD's uncertainty surrounding their results in our simulations, we generally run simulations with a variety of
+output draws.
+
+.. note::
```
kfwilson: extra return here?

docs/source/branch.rst
```diff
@@ -0,0 +1,218 @@
...
+Stochastic Uncertainty
+^^^^^^^^^^^^^^^^^^^^^^
+Vivarium simulations are probabilistic in nature. They tend to contain distributions that describe quantities rather
+than singular values. Because of this, our models are subject to stochastic uncertainty. To attempt to capture some of
+this uncertainty we can run the simulation with different random seeds which will result in different numbers drawn from
+the underlying pseudo-random number generator, and thus different output results. This can be specified in the branches
+file with the ``random_seed_count`` key. This key specifies an integer that represents the number of different random
+seeds to use, each generated randomly and run in a separate simulation.
+
+.. note::
```
kfwilson: I'm just looking at the 'View file' option on github which doesn't render this notes nicely - do they work better with sphinx?

docs/source/branch.rst
```diff
@@ -0,0 +1,218 @@
...
+
+Parameter Variations
+--------------------
+A major function of branches files is to enable easy manipulation of the configuration parameters of a model
+specification to generate different scenarios or examine the sensitivity of a model to changes in a specific parameter.
+In the following sections we will describe a number of ways you can construct different scenarios and the effect on
+the number of simulations these will have.
+
+.. note::
+    The following examples that alter configuration parameters all lie under a ``branches`` key.
```
kfwilson: this note is a little cryptic. Maybe expand a little bit? I think the point you're trying to make is that these changes are under the branches key as opposed to the input draws/seeds which are top level but it's a little unclear whether this note is just saying 'notice the branches key - it's a convenience for the following examples' or is saying 'notice that you have to specify everything except draws & seeds under a 'branches' key. Maybe move the sentence on line 113 about how the branches block changes model spec up to this note?

I also find it a little confusing that we have 'branches key', 'branches file', and 'branches configurations' used in multiple places in this doc. Maybe a clarification somewhere about what these three terms mean?

docs/source/branch.rst
```diff
@@ -0,0 +1,218 @@
...
+            - egg_intervention:
+                    recruitment:
+                        proportion: [0.0, 0.4, 0.8, 1.0]
+                        age_start: [10.0, 25.0, 45.0, 65.0]
+
+This will result in scenarios encompassing every combination of recruitment proportion and starting age. Specifically,
+it will result in 100 separate simulations per combination, one for each randomly sampled draw. This means the total
+number of simulations is given by draws * proportions * starting age, or 1,600 altogether.
+
+.. note:: The top-level keys for controlling seeds and draws are the same across branches.
```
kfwilson: Here again I find the terminology a little confusing - what does 'branches' mean in this note?

docs/source/branch.rst
halfhorst: not until I fix the deps

docs/source/branch.rst
halfhorst: yeah I debated this one. I just wanted to call attention to the difference to try and avoid simple syntax mistakes but maybe it is better left alone

halfhorst: I agree the terminology is all over. what is a configuration, what is a parameter, does a branch include draws and seeds? I'll try and clarify

docs/source/branch.rst
halfhorst: looks like it shows up different on my screen. It is a linewrap to stay under 120. It should render ok

halfhorst:
you can view the docs rendered here: https://vivarium-cluster-tools.readthedocs.io/en/feature-branches-doc/

docs/source/branch.rst
halfhorst: ok, see below

docs/source/branch.rst
kfwilson: hmm the return on line 34 looks like it shows up on the rendered docs though

PR #9, opened 2019-04-10
Add glossary

collijk:
- Added a glossary
- Put in glossary references in the branches text
- Some general edits for clarity
- Follow through with the initial example in the final branches section.

### DIFF
--- a/docs/source/branch.rst
+++ b/docs/source/branch.rst
@@ -74,6 +74,13 @@ but simulation run time scales directly with the size of the population you're s
 you can run multiple simulations with different :term:`random seeds<Random Seed>` and aggregate your results across
 those simulations. This second approach takes advantage of parallel computing to keep run times under control.
 
+.. note::
+   
+    Random seeds are a convenient way to scale up a simulation's population in parallel. For example, running a
+    simulation with one million simulants and a single random seed is equivalent to running the same simulation with
+    ten thousand people and 100 random seeds. Because simulations specified with different seeds will be run
+    in parallel, the latter run strategy is often preferable.   
+
 To run our simulation for multiple random seeds, we use the ``random_seed_count`` key in a
 :term:`branch configuration<Branch Configuration>`. This key specifies an integer that represents the number of
 different random seeds to use, each generated randomly and run in a separate simulation.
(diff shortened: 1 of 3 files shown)


docs/source/branch.rst
```diff
@@ -8,125 +8,167 @@ The Branches File
...
+    produced in sets of 1000 draws.
 
-To do this, we can use the ``input_draw_count`` key in a branches file. This key specifies an integer that represents
-the number of different simulations that will be run in parallel, each with a different draw of GBD results drawn
-uniformly from the number available, which is generally 1,000. Below is a very simple branches file containing only a
-mapping that dictates the number of draws to run.
+To do this, we can use the ``input_draw_count`` key in a :term:`branch configuration<Branch Configuration>`.
+This key refers to an integer that represents the number of different input draws to generate simulations from.
 
 .. code-block:: yaml
```
kfwilson: do these yaml blocks render in sphinx? They totally disappear from the github 'view file' version so I was a bit confused until I came back to the source here

docs/source/branch.rst
```diff
@@ -8,125 +8,167 @@ The Branches File
...
-.. note::
-    Random seeds are a convenient way to scale up a simulation's population in parallel. For example, running a
-    simulation with one million simulants and a single random seed is equivalent to running the same simulation with
-    ten thousand people and 100 random seeds. However, because simulations specified with different seeds will be run
-    in parallel (link psimulate), this is often preferable.
+There are two ways to handle stochastic uncertainty. The first is to increase the size of the population you're
+simulating. This will wash out outlier cases that might heavily skew your results. This works fine up to a point,
+but simulation run time scales directly with the size of the population you're simulating.  Alternatively,
+you can run multiple simulations with different :term:`random seeds<Random Seed>` and aggregate your results across
+those simulations. This second approach takes advantage of parallel computing to keep run times under control.
```
kfwilson: I think this could be spelled out a little more explicitly. It doesn't clearly say that each of the multiple sims can be smaller and thus run faster but that the final aggregation is the same as the big sim. Maybe add back the example comparing a sim of 1m people and 1 seed versus 10k people + 100 random seeds - I'm not sure that this paragraph totally makes it clear that running multiple sims with diff seeds allows essentially splitting your pop.

docs/source/branch.rst
```diff
@@ -8,125 +8,167 @@ The Branches File
...
+we would run ten simulations with the ten random seeds for each of the 100 input data draws.
+
+Configuration Variations
+------------------------
+
+A major function of :term:`branch configurations<Branch Configuration>` is to enable easy manipulation of
+the :term:`configuration parameters<Configuration Parameter>` of a :term:`model specification<Model Specification>`.
+These parameters generally govern interesting features of an intervention, such as it's target coverage or efficacy.
+
+Within a branch configuration, you can specify several variations of these parameters to generate different
```
kfwilson: term reference for this branch configuration?

docs/source/branch.rst
```diff
@@ -8,125 +8,167 @@ The Branches File
...
 
 
 Interaction with Uncertainty
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-As touched upon in the section on :ref:`combining draws and seeds<Combining Draws and Seeds>`, when multiple branch
-configurations would result in multiple simulations, the result is a simulation for every combination, or the cartesian
-product, of the parameters. Let's add draws to our previous branches file and figure out how many simulations it will
-result in.
+As touched upon in the section on :ref:`combining draws and seeds<Combining Draws and Seeds>`, when multiple
+:term:`branch configurations<Branch Configuration>` would result in multiple simulations the result is a simulation for
```
kfwilson: I'm not sure you want branch configurations here - the glossary defines the branch config as the single branches yaml file but this sentence seems to be using it to refer to parameters within that file

docs/source/branch.rst
collijk: Yeah, they do.  A bunch of the directives and stuff will only show up after they render with sphinx

docs/source/branch.rst
halfhorst: yes maybe configuration parameter. Really it is when different configurations would lead to "branches"

docs/source/branch.rst
collijk: I've been using the term reference any the first time a term shows up in a section.  I think even that may be a bit overzealous.

PR #16, opened 2019-04-18
Feature/additional draws

kfwilson:


src/vivarium_cluster_tools/runner.py
```diff
@@ -317,7 +321,7 @@ def check_user_sge_config():
 
 
 def main(model_specification_file, branch_configuration_file, result_directory, project, peak_memory, redis_processes,
-         copy_data=False, num_input_draws=None, num_random_seeds=None, restart=False):
+         copy_data=False, num_input_draws=None, num_random_seeds=None, restart=False, expand={}):
```
halfhorst: :scream:

halfhorst: I don't think it's a problem here, that's just my gut reaction

kfwilson: wait I don't understand. what's causing the reaction? is it bad to have an empty dict default?

halfhorst: mutable defaults can catch ya because the function is setup once by the interpreter, not each time it is called, so if you alter that default all functions get it. I can't see this ever coming up here, though, b/c/ this is CLI code and main isn't gonna be run multiple times

collijk: It's bad practice though.  Swap to a None default.  Good catch.

kfwilson: oh ok good call. changed to None.

PR #21, opened 2019-05-07
cast results to dtypes from pre-np new results

halfhorst:
This preserves type info by hanging on to the dtypes of the fresh results from a worker and using them to cast the results dataframe. Both asking for values from a dataframe and concatenating dataframes of differing types will clobber type info, so we have to do this every time.

For ~1500 columns it adds about a second of overhead per concat_results call


collijk:
Talked with Cody.  To change to a different algorithm that scales better.

PR #23, opened 2019-05-09
Bugfix/mismatched shapes concat

kfwilson:
this finishes the concatenating stuff + fixes another small bug where it tried to write non-existent results if the no_batch flag was turned on. I tested it with a little toy sim (1 cause + 1 risk that runs for 1 year) with 8 random seed/draw combos.

### DIFF
--- a/src/vivarium_cluster_tools/runner.py
+++ b/src/vivarium_cluster_tools/runner.py
@@ -239,7 +239,7 @@ def build_job_list(ctx):
 
 def concat_preserve_types(df_list):
     dtypes = df_list[0].dtypes
-    columns_by_dtype = [list(d[1].index) for d in dtypes.groupby(dtypes)]
+    columns_by_dtype = [list(dtype_group.index) for _, dtype_group in dtypes.groupby(dtypes)]
 
     splits = []
     for columns in columns_by_dtype:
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/runner.py
```diff
@@ -237,27 +237,27 @@ def build_job_list(ctx):
     return jobs
 
 
-def np_concat_preserve_types(df_list):
+def concat_preserve_types(df_list):
     dtypes = df_list[0].dtypes
+    dtypes = [list(d[1].index) for d in dtypes.groupby(dtypes)]
```
collijk: Don't you just need `list(dtypes.unique())` here?

kfwilson: no because I'm grouping by dtypes so I end up with a list of lists where each list contains the columns for a certain dtype

collijk: Oh, nevermind.

Use unpacking though:  ``for _, dtype_group in dtypes.groupby(dtypes)``

kfwilson: but I named that very confusingly. I'll update to a more self-explanatory name

kfwilson: just updated to have a clearer name + use unpacking

PR #26, opened 2019-05-10
Feature/pathlib not os

kfwilson:


src/vivarium_cluster_tools/runner.py
```diff
@@ -138,12 +138,12 @@ def __init__(self, arguments):
         self.peak_memory = arguments.peak_memory
         self.number_already_completed = 0
         self.output_directory = arguments.output_directory
-        self.job_name = Path(arguments.output_directory).resolve().parts[-2]  # The model specification name.
+        self.job_name = arguments.output_directory.parts[-2]  # The model specification name.
```
halfhorst: you can get filename with `stem`, it looks a little cleaner

kfwilson: true but this is actually getting the model spec name from the directory structure. arguments.output_directory is something like `/share/costeffectiveness/results/<MODEL SPEC NAME>/<TIMESTAMP>` and parts[-2] pulls of the model spec name. I think stem would give me the timestamp

PR #27, opened 2019-05-13
handle restarting with only draws/seeds

kfwilson:
hit these two little bugs around restarting/expanding testing the log rotation stuff I set up so this is just the bugs:

1. handle restarting/expanding with no branches (just draws/seeds so branch_config is None not a dict)
2. handle expanding with additional seeds but no draws

### DIFF
--- a/src/vivarium_cluster_tools/runner.py
+++ b/src/vivarium_cluster_tools/runner.py
@@ -195,13 +195,14 @@ def build_job_list(ctx):
         if ctx.existing_outputs is not None:
             mask = ctx.existing_outputs.input_draw == int(input_draw)
             mask &= ctx.existing_outputs.random_seed == int(random_seed)
-            for k, v in collapse_nested_dict(branch_config):
-                if isinstance(v, float):
-                    # FIXME: The fast concat method will coerce floats to objects when writing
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/runner.py
```diff
@@ -195,13 +195,14 @@ def build_job_list(ctx):
...
-                if isinstance(v, float):
-                    # FIXME: The fast concat method will coerce floats to objects when writing
-                    # if there is one or more string columns.  Hack a coercion here to make restart work.
-                    mask &= np.isclose(ctx.existing_outputs[k].astype(float), v)
-                else:
-                    mask &= ctx.existing_outputs[k] == v
+            if branch_config:
+                for k, v in collapse_nested_dict(branch_config):
+                    if isinstance(v, float):
+                        # FIXME: The fast concat method will coerce floats to objects when writing
```
kfwilson: @collijk This FIXME is no longer necessary now that we have the type-preserving concat right? Do we want to leave it in so it's backwards compatible with things we ran before that fix?

collijk: Nope, drop it.

PR #28, opened 2019-05-13
Feature/log rotation

kfwilson:
see image attached of the directory structure created 

image

### DIFF
--- a/src/vivarium_cluster_tools/utilities.py
+++ b/src/vivarium_cluster_tools/utilities.py
@@ -69,11 +69,11 @@ def get_output_directory(model_specification_file=None, output_directory=None, r
 def setup_directories(model_specification_file, result_directory, restart, expand):
     output_directory = get_output_directory(model_specification_file, result_directory, restart)
 
-    if expand:
-        command = 'expand'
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/utilities.py
```diff
@@ -66,6 +66,32 @@ def get_output_directory(model_specification_file=None, output_directory=None, r
     return output_directory
 
 
+def setup_directories(model_specification_file, result_directory, restart, expand):
+    output_directory = get_output_directory(model_specification_file, result_directory, restart)
+
+    if expand:
+        command = 'expand'
+        launch_time = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
+    elif restart:  # expand will also have restart=True so order is important here
```
halfhorst: you could check the pair of booleans to remove the ambiguity. if restart and not expand, if restart and expand. Of course, then we'd be in trouble if we ever pulled apart restart and expand

kfwilson: oh yeah good call. I like removing the ambiguity. changing now

PR #29, opened 2019-05-17
intercept standard logging messages & forward to loguru

kfwilson:


halfhorst:
that's neat. logging is still magic

PR #31, opened 2019-05-31
Feature/version validation

kfwilson:
on first run, saves environment to requirements.txt. on subsequent runs, compares current environment to original. If mismatches, you'll see an error like the following:

image

### DIFF
--- a/src/vivarium_cluster_tools/utilities.py
+++ b/src/vivarium_cluster_tools/utilities.py
@@ -7,6 +7,8 @@ from pathlib import Path
 import sys
 
 from loguru import logger
+from pip._internal.operations import freeze
+from typing import Dict, List, Tuple
 
 from vivarium_cluster_tools import globals as vct_globals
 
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/utilities.py
```diff
@@ -7,6 +7,11 @@
 import sys
 
 from loguru import logger
+from typing import Dict, List, Tuple
+try:
+    from pip._internal.operations import freeze
```
collijk: Document why you need this try/except

src/vivarium_cluster_tools/utilities.py
```diff
@@ -157,3 +162,72 @@ def wrapped(*args, **kwargs):
...
+                         f'In order to successfully run, you should make a new environment using the requirements.txt '
+                         f'file found in the output directory. Differences found as follows: {differences}.')
+
+
+def validate_environment(output_dir: Path):
+    original_environment_file = output_dir / 'requirements.txt'
+
+    current_environment_list = [p for p in freeze.freeze()]
+    if not original_environment_file.exists():  # original run
+        with open(original_environment_file, 'w') as f:
```
collijk: With path.open('w') as f

collijk:
Looks great!  Maybe add some logging?

src/vivarium_cluster_tools/utilities.py
kfwilson: done

src/vivarium_cluster_tools/utilities.py
kfwilson: added comment explaining where freeze is depends on version of pip

PR #32, opened 2019-06-01
Add vadmin command for non simulation utilities

collijk:
This introduces a new cli tool ``vadmin``.  Right now, I just ported over the log parsing stuff @kfwilson just finished.  I'm going to add a tool here for producing new research repos from a template and later (hopefully) an archiving tool.


halfhorst:
this is shaping up nicely as a new click group!

src/vivarium_cluster_tools/vadmin/log_parser.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, file: Path):
         self.task_id = int(name[1])
 
     def _load_messages(self):
-        with open(self.file) as f:
+        with self.file.open('r') as f:
```
halfhorst: is this going to be our preferred method going forward ?

collijk: This is the preferred method when you're hanging onto a pathlib.Path object.  The whole point of the library is object oriented file utilities.

PR #37, opened 2019-06-05
Add better errors, logging, and docs for oauth

collijk:
+ bugfix in oauth_delete_github

### DIFF
--- a/src/vivarium_cluster_tools/vadmin/oauth_utilities.py
+++ b/src/vivarium_cluster_tools/vadmin/oauth_utilities.py
@@ -297,11 +297,12 @@ def oauth_create_github(config: OAuthConfig):
 
 
 def oauth_remove_github(config: OAuthConfig):
-    """Creates a new OAuth token on github."""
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/vadmin/oauth_utilities.py
```diff
@@ -180,23 +282,26 @@ def oauth_remove_stash(config: OAuthConfig):
...
+        'scopes': ['user', 'repo', 'delete_repo']
     }
+    logger.debug('Generating new OAuth token on github.')
     response = requests.post(api_endpoint, auth=user, data=json.dumps(payload))
     token_config = parse_token_creation_response(response)
     config.update('github', token_config)
 
 
 def oauth_remove_github(config: OAuthConfig):
+    """Creates a new OAuth token on github."""
```
kfwilson: I think this should be 'Removes' instead of 'Creates' and do you want a debug msg like in the stash remove method?

PR #38, opened 2019-06-05
Everything up to the cookiecutter bit

collijk:


src/vivarium_cluster_tools/vadmin/repositories.py
```diff
@@ -0,0 +1,230 @@
...
+    Raises
+    ------
+    HTTPError
+        If the repository creation was unsuccessful.
+
+    """
+    if response.status_code == 201:
+        logger.info('Repository successfully created')
+        return response.json()
+    elif response.status_code == 409:
```
kfwilson: this seems odd to raise a HTTPError here - the communication wasn't the cause of the error?

src/vivarium_cluster_tools/vadmin/repositories.py
```diff
@@ -0,0 +1,230 @@
...
+
+    repo_url = create_repository(service, oauth_token, repo_name)
+    try:
+        repo_path = clone_repository(repo_url, output_path)
+    except subprocess.CalledProcessError:
+        delete_repository(service, oauth_token, repo_name)
+        raise
+
+    #
+    # extra = {'project_name': repo_name}
```
halfhorst: Is this supposed to be hanging around ?

collijk: yeah, this is coming in next.

PR #39, opened 2019-06-06
Remaining stuff for research template cloning!

collijk:


src/vivarium_cluster_tools/vadmin/repositories.py
```diff
@@ -220,11 +213,31 @@ def clone_repository(repository_url: str, output_dir: Path) -> Path:
...
+
+    template_url = 'https://github.com/ihmeuw/vivarium_research_template.git'
+    cookiecutter(template_url,
+                 no_input=True,
+                 extra_context=extra,
+                 output_dir=str(repo_path.parent),
+                 overwrite_if_exists=True)
+
+
+def update_repository(repo_path):
```
halfhorst: Can these fail? The first thing I'm thinking of is when a person with unconfigured git tries to push for the first time (though that isn't failure)

The push could fail if internet is down, I suppose

PR #41, opened 2019-06-07
Remove vadmin stuff, name parser command 'vparse'

collijk:
See comment over at https://github.com/ihmeuw/vivarium_development/pull/23


kfwilson:
we no longer need cookiecutter + requests in setup.py right? I was just going to remove them and then realized I wasn't 100% sure (for requests in particular) they weren't used anywhere else

PR #42, opened 2019-07-17
set output directories and log directories to 775

KjellSwedin:


KjellSwedin:
Permissions are relaxed on directories to 775. I didn't make file permission changes although the issue mentioned that. Didn't seem necessary... Not sure I tested restart/expand adequately, although the logging directories show up with "restart" appended to them.

src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -90,10 +98,12 @@ def setup_directories(model_specification_file, result_directory, restart, expan
     logging_dirs = {'main': logging_directory,
                     'sge': logging_directory / 'sge_logs',
                     'worker': logging_directory / 'worker_logs'}
-
+    
     output_directory.mkdir(exist_ok=True, parents=True)
```
collijk: Pathlib let's you set the mode directly when you make the directory: https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir

We might need a separate call to set the group correctly, however. We have a dev group for our team (`ihme-cost-eff-micro`)  that we want to set group ownership with.

KjellSwedin: I implemented with separate calls due to:

_If parents is true, any missing parents of this path are created as needed; they are created with the default permissions without taking mode into account (mimicking the POSIX mkdir -p command)._

kfwilson: hmm I thought you could only change the group to a group that you belong to though. If we set it up so it changes group to our dev group, doesn't that mean that we are the only ones who will be able to run psimulate?

collijk: Makes sense, both.

PR #49, opened 2019-09-11
Feature/add runtime option

halfhorst:
This adds a command line option that maps to the h_rt resource on the cluster. This lets a user run jobs longer than 24 hrs. The cluster default is 24 hours and jobs are reaped at that point.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cli.py
+++ b/src/vivarium_cluster_tools/psimulate/cli.py
@@ -19,8 +19,9 @@ shared_options = [
                  default='24:00:00',
                  help=('The estimated maximum runtime (DD:HH:MM) of the simulation jobs. '
                        'By default, the cluster will terminate jobs after 24h regardless of '
-                       'queue. The maximum supported runtime is 3 days. Keep in mind that runtimes '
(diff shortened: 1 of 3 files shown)


src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -14,6 +14,13 @@
                  default=3,
                  help=('The estimated maximum memory usage in GB of an individual simulate job. '
                        'The simulations will be run with this as a limit.')),
+    click.option('--runtime', '-r',
+                 type=str,
+                 default='24:00:00',
+                 help=('The estimated maximum runtime (DD:HH:MM) of the simulation jobs. '
+                       'By default, the cluster will terminate jobs after 24h regardless of '
+                       'queue. The maximum supported runtime is 3 days. Keep in mind that runtimes '
+                       'by node vary wildly.')),
```
kfwilson: Maybe also mention in this help string that the runtime you've requested for the master process should be at least as long as the runtime requested for each sim job via this flag

collijk:
Actually, Cody, you should do this directly into develop and make a patch version release.

PR #52, opened 2020-01-02
bugfix/determine correct queue based on max runtime

halfhorst:
the cluster has two queues with runtime limits. the runtime dictates the correct queue. we weren't handling long runtimes or invalid runtimes

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/utilities.py
+++ b/src/vivarium_cluster_tools/psimulate/utilities.py
@@ -139,6 +139,17 @@ def get_valid_project(project, cluster):
     return project
 
 
+def get_valid_queue(max_runtime):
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -139,6 +139,17 @@ def get_valid_project(project, cluster):
     return project
 
 
+def get_valid_queue(max_runtime):
+    hours, minutes, seconds = max_runtime.split(":")
```
KjellSwedin: Does something ensure that you have all 3 elements (hours, minutes, seconds)?

halfhorst: we don't. click has support for validation callbacks, this is a good place to start

PR #53, opened 2020-02-18
add queue and priority flags

halfhorst:
Added support for specifying queue and priority, and rearranged some validation / help strings. Priority is validated against cluster-defined limits and queue's default is set dynamically based on runtime.


collijk:
Let's hold this for discussion.  I think we can do some small interface refactoring to clean this up a bit.

PR #55, opened 2020-02-24
Feature/pkg native spec

halfhorst:
this is a potpourri of changes. The functionality is the same, it's just a re-arranging and cleaning up.

the main change is the introduction of the NativeSpecification class to handle all qsub parameters that go into 'nativeSpecification' in the job template and provide validation and a string representation. This means chopping those parameters out in a lot of places.

I also pulled some globals into the globals file, removed a namespace argument because it didn't seem to do anything useful, and remove project validation because click handles that.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cli.py
+++ b/src/vivarium_cluster_tools/psimulate/cli.py
@@ -24,11 +24,6 @@ shared_options = [
                        'queue. The maximum supported runtime is 3 days. Keep in mind that the '
                        'session you are launching from must be able to live at least as long '
                        'as the simulation jobs, and that runtimes by node vary wildly.')),
-    click.option('--threads', '-t',
-                 type=int,
-                 default=1,
-                 help=('The number of threads to request for an individual simulate job. You'
-                       'probably only need one.')),
     click.option('--pdb', 'with_debugger',
                  is_flag=True,
                  help='Drop into python debugger if an error occurs.'),
@@ -93,8 +88,7 @@ def run(model_specification, branch_configuration, result_directory, **options):
     main(model_specification, branch_configuration, result_directory,
          {'project': options['project'],
           'peak_memory': options['peak_memory'],
-          'max_runtime': options['max_runtime'],
-          'threads': options['threads']},
+          'max_runtime': options['max_runtime']},
          redis_processes=options['redis'], no_batch=options['no_batch'])
 
 
@@ -115,8 +109,7 @@ def restart(results_root, **options):
     main(None, None, results_root,
          {'project': options['project'],
           'peak_memory': options['peak_memory'],
-          'max_runtime': options['max_runtime'],
-          'threads': options['threads']},
+          'max_runtime': options['max_runtime']},
          redis_processes=options['redis'], restart=True, no_batch=options['no_batch'])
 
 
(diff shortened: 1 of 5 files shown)


src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -177,6 +178,38 @@ def __init__(self, arguments):
         self.model_specification = self.output_directory / 'model_specification.yaml'
 
 
+# TODO: The dumb thing for now. Can I make this work with NamedTuple ?
+class NativeSpecification:
+
+    def __init__(self, project, peak_memory, max_runtime, validation='n', threads=1, **__):
```
KjellSwedin: Should validation be a boolean?

halfhorst: it's actually a grid engine parameter here with more than two possibilities

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -83,8 +85,8 @@ def run(model_specification, branch_configuration, result_directory, **options):
     utilities.configure_master_process_logging_to_terminal(options['verbose'])
     main = handle_exceptions(runner.main, logger, options['with_debugger'])
 
-    main(model_specification, branch_configuration, result_directory,
-         options['project'], options['peak_memory'], options['max_runtime'],
+    native_specification = runner.NativeSpecification(**options, job_name=Path(model_specification).name)
```
collijk: Do this unpacking inside main.  You're throwing an error that the exception handler can't catch.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -51,7 +50,7 @@ def init_job_template(jt, peak_memory, max_runtime, sge_log_directory, worker_lo
         'SGE_CLUSTER_NAME': sge_cluster,
     }
     jt.joinFiles = True
-    jt.nativeSpecification = utilities.get_uge_specification(peak_memory, max_runtime, project, job_name)
+    jt.nativeSpecification = str(native_specification)
```
collijk: Nice

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -177,6 +175,37 @@ def __init__(self, arguments):
         self.model_specification = self.output_directory / 'model_specification.yaml'
 
 
+class NativeSpecification:
+    def __init__(self, project, peak_memory, max_runtime, validation='n', threads=1, job_name='vivarium', **__):
```
collijk: Defaults should be set in the CLI interface, not here.

src/vivarium_cluster_tools/psimulate/runner.py
KjellSwedin: I figured something like that. If it moves to the CLI for initialization it can become a "choice"

src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -71,13 +69,12 @@ def get_output_directory(model_specification_file=None, output_directory=None, r
         output_directory = root / model_specification_name / launch_time
     return output_directory
 
-DIRECTORY_PERMS=0o775
```
collijk: Instead of pulling this up to globals, make it a function local variable and adjust the docstring to say what it sets.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -177,6 +175,37 @@ def __init__(self, arguments):
...
+        self.project = project
```
collijk: Are we still validating the project somewhere?

halfhorst: the click option validates because its type is choice

PR #56, opened 2020-02-25
typing information

halfhorst:
A bunch of typing 
used magicmock for drmaa instead of object (for typing)
moved RunContext and NativeSpecification to the top (for typing)

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/utilities.py
+++ b/src/vivarium_cluster_tools/psimulate/utilities.py
@@ -27,7 +27,8 @@ def get_drmaa():
                 os.environ['DRMAA_LIBRARY_PATH'] = f'/usr/local/UGE-{sge_cluster_name}/lib/lx-amd64/libdrmaa.so'
             import drmaa
         else:
-            drmaa = object()
+            from unittest.mock import MagicMock
+            drmaa = MagicMock()
     return drmaa
 
 
--- a/src/vivarium_cluster_tools/psimulate/utilities.py
+++ b/src/vivarium_cluster_tools/psimulate/utilities.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from pathlib import Path
 
 from loguru import logger
-from typing import Dict, List, Tuple, Union, Sequence
+from typing import Dict, List, Sequence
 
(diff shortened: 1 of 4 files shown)


src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -78,7 +79,8 @@ def set_permissions(output_dir):
     output_dir.chmod(permissions)
 
 
-def setup_directories(model_specification_file, result_directory, restart, expand):
+def setup_directories(model_specification_file: str, result_directory: str,
+                      restart: bool, expand: bool) -> Tuple[Path, Dict[str, Path]]:
```
collijk: NIT: Think you can just do (Path, Dict[str, Path]) for the return type instead of using `Tuple` explicitly.

src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -52,14 +53,14 @@ def configure_master_process_logging_to_terminal(verbose):
...
-def configure_master_process_logging_to_file(output_directory):
+def configure_master_process_logging_to_file(output_directory: Path):
     master_log = output_directory / 'master.log'
     serial_log = output_directory / 'master.log.json'
     add_logging_sink(master_log, verbose=2)
     add_logging_sink(serial_log, verbose=2, serialize=True)
 
 
-def get_output_directory(model_specification_file=None, output_directory=None, restart=False):
+def get_output_directory(model_specification_file=None, output_directory=None, restart=False) -> Path:
```
collijk: Might as well type the args too.

src/vivarium_cluster_tools/psimulate/utilities.py
```diff
@@ -27,7 +27,8 @@ def get_drmaa():
                 os.environ['DRMAA_LIBRARY_PATH'] = f'/usr/local/UGE-{sge_cluster_name}/lib/lx-amd64/libdrmaa.so'
             import drmaa
         else:
-            drmaa = object()
+            from unittest.mock import MagicMock
```
collijk: Just don't type the session objects

PR #63, opened 2020-04-14
create mkdir utility function

rmudambi:


setup.py
```diff
@@ -47,6 +52,11 @@
         """,
 
         install_requires=install_requires,
+        tests_require=test_requirements,
```
collijk, line 55: Don't do tests_require.  We don't use setup.py to run tests.

PR #75, opened 2021-02-25
MIC-2171: Added worker log timing messages and prototype cli

mattkappel:
This change adds additional timing messages in the worker logs for
setup, simulant initialization, etc., and a final JSON-formatted
telemetry string to be logged for each simulation draw.

This also adds a new tool called "vipin", which is meant to parse
the telemetry strings and other information to produce a csv (or
hdf) of all worker log performance telemetry and log some basic
stats about the psimulate run.

Tested on a full SR Cervical Cancer run v6.0, and produced expected
output.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ if __name__ == "__main__":
             [console_scripts]
             psimulate=vivarium_cluster_tools.psimulate.cli:psimulate
             vparse=vivarium_cluster_tools.vparse.cli:vparse
+            vipin=vivarium_cluster_tools.vipin.cli:vipin
         """,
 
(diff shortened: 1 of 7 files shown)


KjellSwedin (approved):
Nice!

setup.py
```diff
@@ -49,6 +49,7 @@
             [console_scripts]
             psimulate=vivarium_cluster_tools.psimulate.cli:psimulate
             vparse=vivarium_cluster_tools.vparse.cli:vparse
```
collijk, line 51: You can ditch vparse.

mattkappel:
Closed this PR as it was targeted for master

PR #78, opened 2021-03-02
add command line argument to specify artifact path to psimulate

rmudambi:
Tested following scenarios:

Specifying artifact in command line, but not branches file -> Success
Specifying artifact in branches file, but not command line -> Success
Specifying artifact in both command line and branches file -> ConfigurationError

The last commit `1c1d446` fixes a redis `ConnectionError`, but I have no idea why defining `self.model_specification` at the end makes any difference. When I inspected it in the debugger, it was identical before and after the change, but it consistently fails when defining `self.model_specification` at in the block at the beginning of the method and consistently succeeds when defining it at the end.

One other note: writing out the model spec as we are doing here by turning it into a `dict` and then dumping it to yaml is not actually a no-op, though it should be as far as functionality goes. The differences are that all commented lines are lost, tabs are two spaces, while we tend to usually use four spaces, and the order of keys is not preserved.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/branches.py
+++ b/src/vivarium_cluster_tools/psimulate/branches.py
@@ -4,10 +4,12 @@ import yaml
 import numpy as np
 from pathlib import Path
 
+from vivarium.framework.configuration import ConfigurationError
 from vivarium.framework.utilities import collapse_nested_dict
 
+from vivarium_cluster_tools.psimulate import globals as vct_globals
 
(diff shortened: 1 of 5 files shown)


KjellSwedin (approved):

mattkappel (approved):

mattkappel (approved):

collijk:
That redis issue is super weird.

collijk (approved):

PR #79, opened 2021-03-11
Add vipin tool and performance telemetry to parallel worker log

mattkappel:
This PR, based on a previously approved PR, now rebased on the develop branch, adds the vipin tool to parse worker logs and gather Job API data. It also adds new performance telemetry (timings, platform info, etc.) to parallel worker logs.

Tested over multiple LSFF runs on the cluster, data was parsed and csv produced as expected.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ if __name__ == "__main__":
             [console_scripts]
             psimulate=vivarium_cluster_tools.psimulate.cli:psimulate
             vparse=vivarium_cluster_tools.vparse.cli:vparse
+            vipin=vivarium_cluster_tools.vipin.cli:vipin
         """,
 
         install_requires=install_requires,
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,8 @@ if __name__ == "__main__":
         'rq>=1.0, <=1.2.2',
         'vivarium>=0.9.3',
         'click',
+        'psutil',
+        'requests'
     ]
 
     test_requirements = [
--- a/src/vivarium_cluster_tools/psimulate/distributed_worker.py
+++ b/src/vivarium_cluster_tools/psimulate/distributed_worker.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import math
 import os
 from pathlib import Path
@@ -197,7 +198,7 @@ def worker(parameters: Mapping):
         logger.info(f'Total simulation run time {exec_time["total_minutes"]:.3f} minutes.')
 
         # Write out debug JSON line for easy log parsing
-        logger.debug({
+        logger.debug(json.dumps({
             "host": os.environ['HOSTNAME'].split('.')[0],
             "job_number": os.environ['JOB_ID'],
             "task_number": os.environ['SGE_TASK_ID'],
@@ -206,8 +207,8 @@ def worker(parameters: Mapping):
             "scenario": parameters['branch_configuration'],  # assumes leaves of branch config tree are scenarios
             "event": event,
             "exec_time": exec_time,
-            "counters": end_snapshot-start_snapshot
-        })
+            "counters": (end_snapshot-start_snapshot).to_dict()
+        }))
 
         idx = pd.MultiIndex.from_tuples([(input_draw, random_seed)], names=['input_draw_number', 'random_seed'])
         output_metrics = pd.DataFrame(metrics, index=idx)
--- a/src/vivarium_cluster_tools/psimulate/distributed_worker.py
+++ b/src/vivarium_cluster_tools/psimulate/distributed_worker.py
@@ -191,24 +191,7 @@ def worker(parameters: Mapping):
         event['end'] = time()
         end_snapshot = CounterSnapshot()
 
-        exec_time['results_minutes'] = (event['end'] - event["results_start"]) / 60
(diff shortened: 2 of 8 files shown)


setup.py
```diff
@@ -49,6 +51,7 @@
             [console_scripts]
             psimulate=vivarium_cluster_tools.psimulate.cli:psimulate
             vparse=vivarium_cluster_tools.vparse.cli:vparse
```
collijk, line 53: Do this in a separate PR, but we should gut vparse from the code base as well.

src/vivarium_cluster_tools/psimulate/distributed_worker.py
```diff
@@ -155,26 +157,58 @@ def worker(parameters: Mapping):
         logger.info('Simulation configuration:')
         logger.info(str(sim.configuration))
 
-        start = time()
+        start_time = pd.Timestamp(**sim.configuration.time.start.to_dict())
```
collijk, line 160: This all looks like it's doing what it is supposed to, but the performance tracking is heavily interleaved with the computation which really reduces readability.  The first bit here is to make things work, of course, but we should remember to come back here an find a reasonable abstraction for doing this tracking in a way that doesn't obfuscate the computation.  

Three ideas (though there are other ways to do this as well):

- Subclass the SimulationContext and have it do the performance tracking as part of its normal operations
- Subclass the logger to to performance tracking
- Make a PerformanceObserver class that we stick in the sim here and have it track this stuff.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -289,7 +291,7 @@ def build_job_list(ctx: RunContext) -> List[dict]:
 
 def concat_preserve_types(df_list: List[pd.DataFrame]) -> pd.DataFrame:
     """Concatenation preserves all ``numpy`` dtypes but does not preserve any
-    pandas speciifc dtypes (e.g., categories become objects."""
+    pandas specific dtypes (e.g., categories become objects."""
```
collijk, line 294: 👍

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def get_summaries(self) -> dict:
+        """Get all performance summary log messages in WorkerLog"""
+        telemetry_pattern = re.compile(r'^\{\"host\".+\"job_number\".+\}$')
+
+        # Ideally we'd only need to look at the tail of the file, but because a worker can do multiple draws, we need
+        # to iterate over all the lines...
+        # TODO: either change worker to cut a new log for each draw and check tail reversed(list(log))
+        #  *OR* parallelize this
+        f = self.file.open('r')
```
collijk, line 29: As mentioned in the other PR, you want a context manager here.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -0,0 +1,91 @@
+from loguru import logger
```
collijk, line 1: The general convention for sorting python imports is to do them in three blocks:
```
# Stdlib imports
import json
from pathlib import Path
import re
from typing import Union

# Single line of whitespace, then external libs
from loguru import logger
import numpy as np
import pandas as pd
from pandas.io.json import json_normalize
import requests

# Another newline then internal imports
from vivarium_cluster_tools.vipin import utilities  # E.g
```

There are tools we can put in our CI pipeline to keep this sorted for us.

mattkappel: Thanks for this. I'll update for this.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -0,0 +1,91 @@
...
+import numpy as np
+import pandas as pd
+import json
+import re
+from pandas.io.json import json_normalize
+from typing import Union
+import requests
+
+
+class WorkerLog:
```
collijk, line 13: It's not clear that this class is doing a lot of work for you at this point.  Talked about a `to_df` method in the other PR, but maybe a good thing to do here is to write a docstring and describe what the responsibility of this class is.  

I'd expect a class called `WorkerLog` to be a container representing the contents of an on disk log, but what's here is more of a lazy log parser.

mattkappel: WorkerLog was changed to PerformanceLog, but the critique is still the same. I'll add some docstring to help disambiguate what its role is.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -0,0 +1,76 @@
...
+                        random_draw: int = 0):
+    input_directory, output_directory = Path(input_directory), Path(output_directory)
+    worker_log_pattern = re.compile(r'^([0-9]+)\.([0-9]+)\.log$')
+    log_files = [f for f in input_directory.iterdir() if worker_log_pattern.fullmatch(f.name)]
+
+    if random_draw:
+        log_files = sample(log_files, random_draw)
+
+    worker_data = []
+    for i, file in enumerate(log_files):
```
KjellSwedin: Would it make sense to report every 10th, 20th, etc. file as opposed to every file?

KjellSwedin (approved):

rmudambi (approved):

PR #80, opened 2021-03-15
MIC-2307: Add single worker log for all performance log messages

mattkappel:
This change adds a single worker log for all workers to write
the JSON-formatted performance summary log messages to, for
consumption by vipin.

Tested against a real workload (LSFF) with performance.log and
vipin output looking as expected.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/distributed_worker.py
+++ b/src/vivarium_cluster_tools/psimulate/distributed_worker.py
@@ -191,7 +191,7 @@ def worker(parameters: Mapping):
         event['end'] = time()
         end_snapshot = CounterSnapshot()
 
-        do_sim_epilogue(end_snapshot, event, exec_time, parameters, start_snapshot)
+        do_sim_epilogue(start_snapshot, end_snapshot, event, exec_time, parameters)
 
         idx = pd.MultiIndex.from_tuples([(input_draw, random_seed)], names=['input_draw_number', 'random_seed'])
         output_metrics = pd.DataFrame(metrics, index=idx)
@@ -210,7 +210,7 @@ def worker(parameters: Mapping):
         logger.info('Exiting job: {}'.format((input_draw, random_seed, model_specification_file, branch_config)))
 
 
-def do_sim_epilogue(end_snapshot, event, exec_time, parameters, start_snapshot):
+def do_sim_epilogue(start: CounterSnapshot, end: CounterSnapshot, event: dict, exec_time: dict, parameters:dict):
     exec_time['results_minutes'] = (event['end'] - event["results_start"]) / 60
     logger.info(f'Results reporting completed in {exec_time["results_minutes"]:.3f} minutes.')
     exec_time['total_minutes'] = (event['end'] - event["start"]) / 60
@@ -228,6 +228,6 @@ def do_sim_epilogue(end_snapshot, event, exec_time, parameters, start_snapshot):
         "scenario": parameters['branch_configuration'],  # assumes leaves of branch config tree are scenarios
         "event": event,
         "exec_time": exec_time,
-        "counters": (end_snapshot - start_snapshot).to_dict()
+        "counters": (end - start).to_dict()
     }))
     logger.remove(perf_log)
--- a/src/vivarium_cluster_tools/psimulate/distributed_worker.py
+++ b/src/vivarium_cluster_tools/psimulate/distributed_worker.py
@@ -191,24 +191,7 @@ def worker(parameters: Mapping):
         event['end'] = time()
         end_snapshot = CounterSnapshot()
 
-        exec_time['results_minutes'] = (event['end'] - event["results_start"]) / 60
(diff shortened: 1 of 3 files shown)


src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -10,82 +9,68 @@
...
-def parse_log_directory(input_directory: Union[Path, str], output_directory: Union[Path, str], output_hdf: bool,
-                        random_draw: int = 0):
+def parse_log_directory(input_directory: Union[Path, str], output_directory: Union[Path, str], output_hdf: bool):
     input_directory, output_directory = Path(input_directory), Path(output_directory)
-    worker_log_pattern = re.compile(r'^([0-9]+)\.([0-9]+)\.log$')
-    log_files = [f for f in input_directory.iterdir() if worker_log_pattern.fullmatch(f.name)]
-
-    if random_draw:
-        log_files = sample(log_files, random_draw)
+    perf_log = PerformanceLog(Path(input_directory / 'performance.log'))
```
collijk, line 34: `input_directory / 'performance.log'` is a `Path`.  No need to coerce.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -10,82 +9,68 @@
...
 
-    worker_data = []
-    for i, file in enumerate(log_files):
-        # TODO: set up tdqm progress bar
-        logger.info(f'Parsing file {i+1} of {len(log_files)}.')
-        log = WorkerLog(file)
-        for item in log.get_summaries():
-            worker_data.append(item)
-    worker_df = pd.concat(worker_data)
+    perf_data = []
```
collijk, line 36: Why not a `perf_log.to_df()` method?

src/vivarium_cluster_tools/vipin/log_parser.py
mattkappel: Oh good point.

src/vivarium_cluster_tools/vipin/log_parser.py
mattkappel: That's a good suggestion. I'll change that.

src/vivarium_cluster_tools/psimulate/distributed_worker.py
```diff
@@ -225,3 +208,26 @@ def worker(parameters: Mapping):
         raise
     finally:
         logger.info('Exiting job: {}'.format((input_draw, random_seed, model_specification_file, branch_config)))
+
+
+def do_sim_epilogue(end_snapshot, event, exec_time, parameters, start_snapshot):
```
collijk, line 213: Typing on the args here would be super helpful.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -10,82 +9,68 @@
...
-        name = file.stem.split('.')  # TODO: validate name is <sge job id>.<task id>.log
-        self.sge_job_id = int(name[0])
-        self.task_id = int(name[1])
+        self.count = 0
 
     def get_summaries(self) -> dict:
-        """Get all performance summary log messages in WorkerLog"""
-        telemetry_pattern = re.compile(r'^\{\"host\".+\"job_number\".+\}$')
+        """Get all performance summary log messages in PerformanceLog"""
+        telemetry_pattern = re.compile(r'^{\"host\".+\"job_number\".+}$')
```
collijk, line 19: You're telemetry pattern is probably better as a class constant.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -10,82 +9,68 @@
...
 
-        # Ideally we'd only need to look at the tail of the file, but because a worker can do multiple draws, we need
-        # to iterate over all the lines...
-        # TODO: either change worker to cut a new log for each draw and check tail reversed(list(log))
-        #  *OR* parallelize this
+        self.count = 0
```
collijk, line 21: Count, on the other hand, looks like it should be a local variable.

mattkappel: There isn't strong purpose for the count, it's unused. My original intent was to report the number of entries, but the count is redundant to, say, df.shape. I'll remove it.

src/vivarium_cluster_tools/vipin/log_parser.py
```diff
@@ -10,82 +9,68 @@
...
         f = self.file.open('r')
```
collijk, line 22: You always want to use a context manager pattern for file opening (and pretty much every other situation that looks like 1. acquire resource, 2. use resource, 3. free resource).  

```
with self.file.open() as f:  # Will default open for read
   ... do stuff with f ...
# Exiting the block guarantees things are cleaned up
```

collijk (approved):
Overall looks good!

PR #85, opened 2021-04-02
Feature/mic 2308 2320 automate vipin delete vparse

mattkappel:
This change adds a call in the parallel runner to
automatically run vipin after processing results. 
It also removes vparse from the codebase in
favor of replacement with vipin.

There are also a few minor aesthetic changes: 
whitespace changes for PEP-8, refactoring
of the atexit-called method to remove the results
directory on no results being produced, and
added a docstring for the same.

Tested with parallel runs of LSFF and Lung 
Cancer simulations.


KjellSwedin (approved):

mattkappel:
On testing, it's not clean in the error/no-results case. I'm going to make a new PR after testing.

PR #104, opened 2021-12-10
Change pip freeze usage to use command rather than internal method

mattkappel:
This changes the `pip freeze` usage from an internal method to the command. This allows us to use the `--format` flag that in testing removes the issue in MIC-2580, where local file paths are used instead of versions.

Tested with multiple myeloma run, killed after producing the `requirements.txt`. Tested also with a restart, killed after passing the environment validation step.


mattkappel:
There are some ancillary commits from `main` because I originally based off of `main` instead of `develop`... The changes at hand (non-release related) are limited to `src/vivarium_cluster_tools/psimulate/utilities.py`

hussain-jafari (approved):

rmudambi (approved):

collijk (approved):

PR #107, opened 2022-02-07
log error if batching retries run out

rmudambi:
Adding logging of error while we decide on a long-term solution

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/runner.py
+++ b/src/vivarium_cluster_tools/psimulate/runner.py
@@ -348,10 +348,12 @@ def write_results_batch(ctx: RunContext,
             results_to_write.to_hdf(temp_output_path, 'data')
             temp_output_path.replace(output_path)
             break
-        except Exception:
+        except Exception as e:
             logger.warning(f'Error trying to write results to hdf, retries remaining {retries}')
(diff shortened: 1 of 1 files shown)


albrja (approved):

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -348,10 +348,12 @@ def write_results_batch(ctx: RunContext,
             results_to_write.to_hdf(temp_output_path, 'data')
             temp_output_path.replace(output_path)
             break
-        except Exception:
+        except Exception as e:
             logger.warning(f'Error trying to write results to hdf, retries remaining {retries}')
             sleep(30)
             retries -= 1
+            if not retries:
+                logger.warning(f"Retries exhausted: {e}")
```
collijk, line 356: I would log a warning here that says the filesystem is unreachable, and then actually raise the error.  Otherwise, you're continuing your psimulate run even though you have failed to write a batch of results, or something bad has happened when you're replacing the existing file, or some some other i/o related shenanigans. You're also very likely in a place (given that you've already done retries) where the next batch of results is going to have the same errors.

collijk (approved):

PR #135, opened 2022-02-20
Feature/load test

collijk:
### Description

- *Category*: feature
- *JIRA issue*: [MIC-2357](https://jira.ihme.washington.edu/browse/MIC-2357)

This adds a new `test` command to psimulate that takes a `test_type` argument 
("sleep" is the only current valid option). Example usage:
```
psimulate test sleep -n 10 -vv --pdb   
```
which will launch ten jobs that sleep for a random amount of time between 5 and 10
seconds (the test_types themselves are not parameterizable outside changes
in the code right now).  

Because this is the feature that also makes integration testing possible, 
it also has all the bugfixes from the bugs I introduced in the refactor.  Sorry!

I'm playing a little fast and loose here. Things need docstrings and I'm passing 
around or relying on some raw string data. There are also several more kinds 
of tests to add and then the corresponding fixes we can make with the results
of the tests.  I'm going to delay fixing those issues for a bit while I do integration 
testing with ciff and  mnch and then pivot to the slurm transition.  

### Testing
This is the test!


albrja (approved):

mattkappel (approved):

src/vivarium_cluster_tools/psimulate/worker/load_test_work_horse.py
```diff
@@ -0,0 +1,60 @@
...
+        logger.exception("Unhandled exception in worker")
+        job = get_current_job()
+        job.meta["root_exception"] = format_exc()
+        job.save_meta()
+        raise
+    finally:
+        logger.info(f"Exiting job: {job_parameters}")
+
+
+def sleep_test(job_parameters: JobParameters) -> pd.DataFrame:
```
beatrixkh, line 49: So this is #1 in the ticket's listed test types, and we'll open other tickets for some of these other examples?

```
1. Start a job, sleep for passed in # seconds (as a parameter); Return an empty data frame 
2 .Starts and generates a numpyarray (to consume memory); Returns an empty data frame if successful
3. Starts and generates a data frame with large qty of long column names and returns them (something that looks like multiple myeloma) - "SwampMonster"
Others
```

PR #138, opened 2022-03-24
use artifact to generate results directory name

rmudambi:
### Description

- *Category*: bugfix
- *JIRA issue*: [MIC-2969](https://jira.ihme.washington.edu/browse/MIC-2969)

Use the artifact path to determine output directory name rather than the model-spec path.

### Testing
Ran psimulate and saw that the output directory was named as expected

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/paths.py
+++ b/src/vivarium_cluster_tools/psimulate/paths.py
@@ -70,7 +70,7 @@ class OutputPaths(NamedTuple):
         cls,
         *,
         command: str,
-        input_model_specification_path: Optional[Path],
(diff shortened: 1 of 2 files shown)


albrja (approved):

mattkappel:
What is the argument for making this change? I presume it's uncontroversial because both names are usually the same, right? Still, it seems unintuitive to have output path depend on the artifact path.

rmudambi:
Currently they don't go into location-specific directories, as there is only one model spec file for a given model, even if we run across multiple locations. It makes sense for the directories to match the input data, as that is the differentiator.

PR #146, opened 2022-06-23
Make draw and seed generation deterministic in parallel simulations

mattkappel:
### Description
- *Category*: bugfix
- *JIRA issue*: [MIC-3193](https://jira.ihme.washington.edu/browse/MIC-3193)

#### Changes
- Uses `np.random.shuffle` rather than `np.random.choice` to choose the requested amount of draws and seeds
- Note that this sets the max seed count to 10000. This is presumably reasonable.

### Testing
Ran `psimulate run` and `psimulate expand` with breakpoints to watch draws and seeds be drawn deterministically.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/branches.py
+++ b/src/vivarium_cluster_tools/psimulate/branches.py
@@ -138,6 +138,8 @@ def calculate_input_draws(
         min_input_draw_count_allowed = 1
     np.random.shuffle(possible)
     if min_input_draw_count_allowed <= input_draw_count <= len(possible):
+        # XXX
+        breakpoint()
         return possible[:input_draw_count]
     else:
         raise ValueError(
@@ -176,9 +178,9 @@ def calculate_random_seeds(
 
     if existing_seeds:
         possible = list(set(possible).difference(existing_seeds))
+    np.random.shuffle(possible)
     # XXX
     breakpoint()
-    np.random.shuffle(possible)
     return possible[:random_seed_count]
 
 
--- a/src/vivarium_cluster_tools/psimulate/branches.py
+++ b/src/vivarium_cluster_tools/psimulate/branches.py
@@ -132,12 +132,10 @@ def calculate_input_draws(
     np.random.seed(123456)
     possible = list(range(1000))
     if existing_draws:
-        possible = [list(set(possible).difference(existing_draws))]
+        possible = list(set(possible).difference(existing_draws))
         min_input_draw_count_allowed = 0
     else:
         min_input_draw_count_allowed = 1
-    # XXX
-    breakpoint()
     np.random.shuffle(possible)
     if min_input_draw_count_allowed <= input_draw_count <= len(possible):
         return possible[:input_draw_count]
@@ -177,7 +175,7 @@ def calculate_random_seeds(
     possible = list(range(100000))
 
     if existing_seeds:
-        possible = [list(set(possible).difference(existing_seeds))]
+        possible = list(set(possible).difference(existing_seeds))
     # XXX
     breakpoint()
     np.random.shuffle(possible)
--- a/src/vivarium_cluster_tools/psimulate/branches.py
+++ b/src/vivarium_cluster_tools/psimulate/branches.py
@@ -129,8 +129,12 @@ def calculate_input_draws(
         existing draw numbers.
 
     """
+    MAX_DRAW_COUNT = 1000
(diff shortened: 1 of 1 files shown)


src/vivarium_cluster_tools/psimulate/branches.py
```diff
@@ -129,16 +129,20 @@ def calculate_input_draws(
...
+    possible = list(range(MAX_DRAW_COUNT))
     if existing_draws:
-        possible = list(set(range(1000)).difference(existing_draws))
+        possible = list(set(possible).difference(existing_draws))
         min_input_draw_count_allowed = 0
     else:
-        possible = list(range(1000))
         min_input_draw_count_allowed = 1
-
+    np.random.shuffle(possible)
```
rmudambi, line 143: I'd suggest moving the seed call closer to the shuffle.

src/vivarium_cluster_tools/psimulate/branches.py
```diff
@@ -129,16 +129,20 @@ def calculate_input_draws(
...
     """
+    MAX_DRAW_COUNT = 1000
+    if input_draw_count > MAX_DRAW_COUNT:
+        raise ValueError(f"Input draw count must be less than {MAX_DRAW_COUNT}.")
+
     np.random.seed(123456)
+    possible = list(range(MAX_DRAW_COUNT))
     if existing_draws:
-        possible = list(set(range(1000)).difference(existing_draws))
+        possible = list(set(possible).difference(existing_draws))
```
rmudambi, line 139: Do we have guarantees that the set difference operation we do here will always give the same ordering of the resultant set? It might be worth avoiding sets altogether and just doing the slower list operations since it's only 1000 draws and we're only doing it once.

src/vivarium_cluster_tools/psimulate/branches.py
```diff
@@ -171,17 +175,16 @@ def calculate_random_seeds(
...
     if existing_seeds:
-        min_possible = max(existing_seeds) + 1
-    else:
-        existing_seeds = []
-        min_possible = 0
-
-    low, high = min_possible, min_possible + 10 * random_seed_count
-    possible = list(range(low, high))
-    return np.random.choice(possible, random_seed_count, replace=False).tolist()
+        possible = list(set(possible).difference(existing_seeds))
```
rmudambi, line 185: Same as above about sets vs lists.

rmudambi (approved):

albrja (approved):

src/vivarium_cluster_tools/psimulate/branches.py
```diff
@@ -129,16 +129,20 @@ def calculate_input_draws(
         existing draw numbers.
 
     """
+    MAX_DRAW_COUNT = 1000
```
stevebachmeier, line 132: There's a near-zero chance this would ever change at IHME, right? Otherwise is there a centralized globals.py or something similar this can be stored?

src/vivarium_cluster_tools/psimulate/branches.py
stevebachmeier: Or you can `sorted(...)`

stevebachmeier: Also, TIL that you can `set.difference(list)`

src/vivarium_cluster_tools/psimulate/branches.py
rmudambi: I'm not aware of a central location where constants like this are defined. If this did change it would presumably be accompanied by some sort of notice, so we could change it with them. I suspect it would screw with a lot of things if that changed, but I don't think it's worth worrying about right now.

src/vivarium_cluster_tools/psimulate/branches.py
mattkappel: I don't think this is an issue. When we do this operation, possible is ordered numerically increasing. The result of the difference is similarly ordered, minus those draws in the existing draws.

mattkappel: I added the sorted call, even though it's likely superfluous.

src/vivarium_cluster_tools/psimulate/branches.py
mattkappel: I'm going to keep it as sets, with the sorted call. In terms of performance, there didn't seem to be a significant difference.

PR #150, opened 2022-08-08
 Enable `psimulate` to run without artifact

mattkappel:
### Description
- *Category*: bugfix
- *JIRA issue*:  [MIC-3229](https://jira.ihme.washington.edu/browse/MIC-3229) and [MIC-3091](https://jira.ihme.washington.edu/browse/MIC-3091)

Changes:
- Allow `psimulate` to run without an artifact.
- If no artifact is provided, use the model spec yaml's stem to be the stand-in for location in results path.

### Testing
Ran this VCT against IV Iron with an artifact provided only in the YAML and only at the command and @aflaxman's WA DOH staffing repo, which does not have an artifact and motivated one of these JIRA tickets. In each case, the correct artifact (if any) and results path location were as expected (`"south_asia"` in the former case and `"staffing"` in the latter, from `staffing.yaml`

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/paths.py
+++ b/src/vivarium_cluster_tools/psimulate/paths.py
@@ -91,7 +91,7 @@ class OutputPaths(NamedTuple):
                 with open(input_model_spec_path) as model_spec_file:
                     model_spec = yaml.safe_load(model_spec_file)
                 try:
-                    location = model_spec["configuration"]["input_data"]["artifact_path"].stem
+                    location = Path(model_spec["configuration"]["input_data"]["artifact_path"]).stem
                 except KeyError:
                     location = input_model_spec_path.stem
             output_directory = output_directory / location / launch_time
--- a/src/vivarium_cluster_tools/psimulate/paths.py
+++ b/src/vivarium_cluster_tools/psimulate/paths.py
@@ -8,6 +8,8 @@ from datetime import datetime
 from pathlib import Path
 from typing import NamedTuple, Optional, Union
 
+import yaml
+
 from vivarium_cluster_tools import utilities as vct_utils
 from vivarium_cluster_tools.psimulate import COMMANDS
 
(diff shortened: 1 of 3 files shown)


src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -70,14 +72,29 @@ def from_entry_point_args(
...
+            # model_specification.py code.
+            if input_artifact_path:
+                location = input_artifact_path.stem
+            else:
+                # check for YAML
+                with open(input_model_spec_path) as model_spec_file:
+                    model_spec = yaml.safe_load(model_spec_file)
+                try:
+                    location = Path(model_spec["configuration"]["input_data"]["artifact_path"]).stem
+                except KeyError:
```
stevebachmeier, line 95: This kind of feels borderline like the type of error you shouldn't catch b/c it's expected behavior (please correct me if I'm wrong here). Regardless, another approach I defaulted to when thinking about this was implementing .get(). eg

```
location = model_spec.get('configuration', {}).get('input_data', {}).get('artifact_path')
if not location:
    location = input_model_spec_path.stem
```

Or I guess if you wanted to golf:
```
location = model_spec.get('configuration', {}).get('input_data', {}).get('artifact_path', input_model_spec_path.stem)
```

src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -70,14 +72,29 @@ def from_entry_point_args(
...
+            # which must exist. NB: Actual determination of the artifact path is done in the
+            # model_specification.py code.
+            if input_artifact_path:
+                location = input_artifact_path.stem
+            else:
+                # check for YAML
+                with open(input_model_spec_path) as model_spec_file:
+                    model_spec = yaml.safe_load(model_spec_file)
+                try:
+                    location = Path(model_spec["configuration"]["input_data"]["artifact_path"]).stem
```
rmudambi, line 94: In what context would `input_artifact_path` not exist, but there be an artifact path defined in the model spec file? This seems like it should never happen.

mattkappel: None, but you still have to look. Three possibilities:

CLI
YAML
none

In the first two we use the artifact name, in the latter we use the name of the YAML file.

mattkappel: Oh, you mean why specify artifact in the YAML but not the CLI? 

I prefer it tbqh. Fewer options to pass to psimulate.

src/vivarium_cluster_tools/psimulate/paths.py
mattkappel: I hesitated on the try/except because of that, it it fits so well. Do you think your code is clearer than the OP?

stevebachmeier: I think it's clearer and results in one less indent. But I don't think it's a big deal either way

mattkappel: OTOH, it doesn't seem like it's bad form in Python. EAFP.

mattkappel: https://stackoverflow.com/questions/11360858/what-is-the-eafp-principle-in-python

stevebachmeier: The second comment on that thread actually points out the .get() approach. Seems like it doesn't matter much. I prefer minimizing indents (and .get() is readable to me), but what you have is fine as well.

mattkappel: Looking at efficiency, the EAFP approach reads the dictionary once. The `.get()` approach reads it three times.

albrja (approved):

mattkappel:
I'm going to take a different tack after chatting with @rmudambi to have `simulate` and `psimulate` use the same code to choose the location-like name used in the results directory. Right now, `simulate` doesn't infer from artifact, it only takes the stem of the model spec yaml. The code will be effectively the same as presented here, but moved/refactored into the `vivarium` repo

PR #151, opened 2022-08-10
Infer location from artifact, if possible, otherwise use model spec

mattkappel:
### Description
- *Category*: feature
- *JIRA issue*:  [MIC-3229](https://jira.ihme.washington.edu/browse/MIC-3229) and [MIC-3091](https://jira.ihme.washington.edu/browse/MIC-3091)

Changes:
- Allow `psimulate` to run without an artifact.
- If no artifact is provided, use the model spec yaml's stem to be the stand-in for location in results path.
- If no artifact is provided, emit a debug message asking "are you sure?"
- This change relies on a change in `vivarium` to include `get_output_model_name_string()` (https://github.com/ihmeuw/vivarium/pull/220) 

### Testing
Ran this VCT and Vivarium against IV Iron with an artifact provided only in the YAML and only at the command and @aflaxman's WA DOH staffing repo, which does not have an artifact and motivated one of these JIRA tickets. In each case, the correct artifact (if any) and results path location were as expected (`"south_asia"` in the former case and `"staffing"` in the latter, from `staffing.yaml`

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/paths.py
+++ b/src/vivarium_cluster_tools/psimulate/paths.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import NamedTuple, Optional, Union
 
+from vivarium.interface.utilities import get_output_location_like_string
 from vivarium_cluster_tools import utilities as vct_utils
 from vivarium_cluster_tools.psimulate import COMMANDS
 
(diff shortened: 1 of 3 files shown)


src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -72,12 +73,14 @@ def from_entry_point_args(
...
         input_artifact_path: Optional[Path],
         result_directory: Path,
+        input_model_spec_path: Path,
     ) -> "OutputPaths":
         launch_time = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
 
         output_directory = result_directory
         if command == COMMANDS.run:
-            output_directory = output_directory / input_artifact_path.stem / launch_time
+            location = get_output_location_like_string(input_artifact_path, input_model_spec_path)
```
collijk, line 82: Here too, there's no real reason to document that this is a location.  You can use this function (renamed to something more sensible like `get_model_name`) without making reference to the fact that it's a location.  E.g. PRL artifacts may not end up being named after locations, but after things like `census` and `cancer_registry` as we use vivarium to generate synthetic data for record linkage.

albrja (approved):

rmudambi (approved):

mattkappel:
Oops, I merged to `main`. I've fixed that mistake.

mattkappel:
It doesn't look like I can re-direct this PR at `develop` I've created https://github.com/ihmeuw/vivarium_cluster_tools/pull/152 to merge to `develop`.

PR #171, opened 2023-06-13
Allow specification of random seeds in branches file

mattkappel:
### Description
- *Category*: feature
- *JIRA issue*: [MIC-4107](https://jira.ihme.washington.edu/browse/MIC-4107)

### Changes and notes
- Adds random_seeds key for listing specific random_seeds, similar to the input_draws key.
- Refactors validation of input_draws to be leveraged with random_seeds validation.

### Testing
Tested manually against the IV Iron maternal simulation. I got the expected behavior in the normal case and two failure cases (random_seed_count != len(random_seeds) and if any value in random_seeds is outside of the range (0-9999).

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/branches.py
+++ b/src/vivarium_cluster_tools/psimulate/branches.py
@@ -274,7 +274,7 @@ def _check_count_and_values(
     if values:
         if [d for d in values if d not in range(0, max_count)]:
             raise ValueError(
-                f"{values} contains draws outside of 0-{max_count - 1}: "
+                f"{values_name} contains draws outside of 0-{max_count - 1}: "
                 f"{[d for d in values if d not in range(0, max_count)]}"
             )
     if value_count < 1 or value_count > max_count:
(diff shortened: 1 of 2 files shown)


hussain-jafari (approved):

src/vivarium_cluster_tools/psimulate/branches.py
```diff
@@ -206,34 +209,76 @@ def calculate_keyspace(branches: List[Dict]) -> Dict:
...
+    _check_count_and_values(
+        data, input_draw_count, input_draws, "input_draw_count", "input_draws", 1000
+    )
+    _check_count_and_values(
+        data,
+        random_seed_count,
+        random_seeds,
+        "random_seed_count",
+        "random_seeds",
+        10000,
```
stevebachmeier, line 232: Where does this number come from? Just a reasonable max?

stevebachmeier (approved):

src/vivarium_cluster_tools/psimulate/branches.py
mattkappel: Good question! For this PR, it comes from the existing seeds implementation. That said, I don't think there is a strong constraint on the max number of seeds or that they must be between 0-9999.

PR #173, opened 2023-06-23
Automate versioning using setuptools_scm

mattkappel:
### Description
- *Category*: POC
- *JIRA issue*: [MIC-4167](https://jira.ihme.washington.edu/browse/MIC-4167)

### Changes and notes
Adds usage of setuptools_scm. Most of the defaults work for our use case:
* Releases where there are no changes between the latest version tag will get the version and no other information. This is PyPI compliant.
* When we have revisions after the tag, or the work area is "dirty", you'll get the distance (number of revisions) and the latest git commit, e.g, 1.3.10.dev2+g1de891c, where the last tag is v1.3.9, 2 commits since that tag and the latest commit is 1de891c.

The `use_scm_version` block in `setup.py` is taken from [an example in the documentation](https://github.com/pypa/setuptools_scm#configuration-parameters). `use_scm_version` says to use setuptools scm, `write_to` and `write_to_template` manages the `_version.py` file writing. The `tag_regex` matches tags of valid version names, used in discerning the current version. This regex could be paired back, since we'll never release a version that has a `+` segment (PyPI disallows this AFAIK), but it's not necessary, since picking the version tag is still (and intentionally) a manual process.

### Testing
Works manually. Tested with `python setup.py --version` and `pip install`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,6 @@ import os
 from setuptools import find_packages, setup
 
 if __name__ == "__main__":
-
     base_dir = os.path.dirname(__file__)
     src_dir = os.path.join(base_dir, "src")
 
@@ -29,9 +28,7 @@ if __name__ == "__main__":
         "requests",
     ]
 
-    setup_requires = [
-        "setuptools_scm"
-    ]
+    setup_requires = ["setuptools_scm"]
 
     test_requirements = [
         "pytest",
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,10 @@ if __name__ == "__main__":
         "requests",
     ]
 
+    setup_requires = [
+        "setuptools_scm"
+    ]
+
     test_requirements = [
         "pytest",
         "pytest-mock",
@@ -44,7 +48,6 @@ if __name__ == "__main__":
 
     setup(
         name=about["__title__"],
-        version=about["__version__"],
         description=about["__summary__"],
         long_description=long_description,
         url=about["__uri__"],
@@ -65,4 +68,10 @@ if __name__ == "__main__":
             "dev": doc_requirements + test_requirements,
         },
         zip_safe=False,
+        use_scm_version={
+            "write_to": "_version.py",
+            "write_to_template": '__version__ = "{version}"',
+            "tag_regex": r"^(?P<prefix>v)?(?P<version>[^\+]+)(?P<suffix>.*)?$",
+        },
+        setup_requires=setup_requires,
     )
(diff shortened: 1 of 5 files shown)


mattkappel:
I see the doc build is failing, I will mend that.

setup.py
```diff
@@ -65,4 +65,10 @@
             "dev": doc_requirements + test_requirements,
         },
         zip_safe=False,
+        use_scm_version={
+            "write_to": "_version.py",
+            "write_to_template": '__version__ = "{version}"',
+            "tag_regex": r"^(?P<prefix>v)?(?P<version>[^\+]+)(?P<suffix>.*)?$",
```
rmudambi, line 71: Can you walk me through what this is doing?

mattkappel: The `use_scm_version` block in `setup.py` is taken from [an example in the documentation](https://github.com/pypa/setuptools_scm#configuration-parameters). `use_scm_version` says to use setuptools scm, `write_to` and `write_to_template` manages the `_version.py` file writing. The `tag_regex` matches tags of valid version names, used in discerning the current version. This regex could be paired back on retrospect, since we'll never release a version that has a `+` segment (PyPI disallows this AFAIK)

mattkappel:
All right, this PR is officially ready for review!

rmudambi (approved):
It'd be great if you presented to the team what this is doing. Could be a good topic for Friday's discussion. Assuming this works, it looks good to me.

setup.py
```diff
@@ -65,4 +65,10 @@
             "dev": doc_requirements + test_requirements,
         },
         zip_safe=False,
+        use_scm_version={
+            "write_to": "src/vivarium_cluster_tools/_version.py",
+            "write_to_template": '__version__ = "{version}"\n',
+            "tag_regex": r"^(?P<prefix>v)?(?P<version>[^\+]+)(?P<suffix>.*)?$",
```
stevebachmeier, line 71: And this regex is well-tested and/or well-known?

setup.py
```diff
@@ -65,4 +65,10 @@
             "dev": doc_requirements + test_requirements,
         },
         zip_safe=False,
+        use_scm_version={
+            "write_to": "src/vivarium_cluster_tools/_version.py",
+            "write_to_template": '__version__ = "{version}"\n',
```
stevebachmeier, line 70: Is this supposed to be an f-string? Where does `{version}` come from?

stevebachmeier (approved):

setup.py
mattkappel: It's in the setuptools_scm README as an example.

setup.py
mattkappel: No, not an f-string here. This is for setuptools_ACM to use to populate _version.py at setup.

hussain-jafari (approved):

PR #185, opened 2023-09-28
Add Max Workers option to throttle jobs

patricktnast:
##Add Max Workers option to throttle jobs

### Description

- *Category*: 
feature
- *JIRA issue*: [MIC-4554](https://jira.ihme.washington.edu/browse/MIC-4554)

### Changes and notes
 
Add a click option to set max number of workers/jobs to submit to the cluster.
Since the workers are created in "burst" mode, they will automatically pick up workers from their queue as they finish.

A failure mode is that some care should be taken to ensure that there are many more workers than queues, so that no queue is left entirely empty.
### Testing

Tested basic functionality with MNCH:
--max jobs option does throttle number of concurrent jobs
--warning is given if you specify too few workers per queue
--0 input produces an error
--having a very high threshold gives you the default, total number of jobs.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/redis_dbs/launcher.py
+++ b/src/vivarium_cluster_tools/psimulate/redis_dbs/launcher.py
@@ -20,15 +20,15 @@ from loguru import logger
 from vivarium_cluster_tools.psimulate.environment import ENV_VARIABLES
 
 DEFAULT_NUM_REDIS_DBS = -1
-DEFAULT_JOBS_PER_REDIS_INSTANCE = 1000
+DEFAULT_WORKERS_PER_REDIS_INSTANCE = 1000
 
 
 def launch_redis_processes(
     num_processes: int,
-    num_jobs: int,
+    num_workers: int,
     redis_logging_root: Path,
 ) -> List[Tuple[str, int]]:
-    num_processes = _get_num_redis_dbs(num_processes, num_jobs)
+    num_processes = _get_num_redis_dbs(num_processes, num_workers)
 
     hostname = ENV_VARIABLES.HOSTNAME.value
     redis_ports = []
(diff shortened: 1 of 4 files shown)


patricktnast:
regarding the vulnerability above, it should be pretty unlikely under normal psimulate conditions
https://en.wikipedia.org/wiki/Coupon_collector%27s_problem

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -193,11 +194,17 @@ def main(
     else:
         logger.info(f"Found {len(job_parameters)} jobs to run.")
 
+    num_workers = (
+        extra_args["max_workers"]
+        and min(extra_args["max_workers"], len(job_parameters))
+        or len(job_parameters)
+    )
+
     logger.info("Spinning up Redis DBs and connecting to main process.")
```
mattkappel, line 203: Can you explain the logic here?

patricktnast: if max_workers is truthy, return min (max_workers, num_jobs)
else return num_jobs
this had the kind of nice property of evaluating to num_jobs if 0 was passed into max_workers, but in the end I just restricted that through the click option
On reflection, it's probably not best to use the idiomatic and/or ternary construction

mattkappel: Nvm, I see you changed the expression completely.

src/vivarium_cluster_tools/psimulate/redis_dbs/launcher.py
```diff
@@ -89,3 +98,9 @@ def _get_random_free_port() -> int:
     port = s.getsockname()[1]
     s.close()
     return port
+
+
+def _expected_sufficient_workers(num_queues) -> int:
+    # Rough estimate of the number of workers needed to ensure that each queue gets
+    # at least one worker. cf. https://w.wiki/7bnb
+    return int(math.ceil(num_queues * (math.log(num_queues) + 0.57)))
```
rmudambi, line 106: What happens if you specify 1 worker and 1 queue?

src/vivarium_cluster_tools/psimulate/runner.py
rmudambi: Could alternately do this:
```python
num_workers = len(job_parameters)
if extra_args["max_workers"]:
    num_workers = min(extra_args["max_workers"], num_workers)
```

rmudambi (approved):

PR #187, opened 2023-10-02
Fix Job Params

patricktnast:
### Description

- *Category*: 
bugfix
- *JIRA issue*: [MIC-4332](https://jira.ihme.washington.edu/browse/MIC-4332)

### Changes and notes
 
It turns out we are depended on the JobParams object itself having its branch config updated, so updating a copy as a ConfigTree left things out of the results .hdf

It is also not so easy to just use the sim.configuration tree in place of job_parameters, since theres a lot of extra stuff. In the end, i went with a minimal change although this preserves the (somewhat more explicit now) dependence on job parameters being updated. I think it would be a good idea to consider combing through this for a more extensive refactor; i think we could probably streamline this a bit
### Testing

added a bit more test coverage. Tested with MNCH maternal model

### DIFF
--- a/tests/psimulate/worker/test_vivarium_work_horse.py
+++ b/tests/psimulate/worker/test_vivarium_work_horse.py
@@ -1,11 +1,12 @@
-import pytest
-
 from time import time
+
+import pytest
 from vivarium.framework.utilities import collapse_nested_dict
+
 from vivarium_cluster_tools.psimulate.jobs import JobParameters
 from vivarium_cluster_tools.psimulate.worker.vivarium_work_horse import (
-    setup_sim,
     parameter_update_format,
+    setup_sim,
 )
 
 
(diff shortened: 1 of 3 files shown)


mattkappel (approved):

tests/psimulate/worker/test_vivarium_work_horse.py
```diff
@@ -21,7 +30,19 @@ def test_setup_sim(mocker):
...
-    sim_config = setup_sim(job_parameters).configuration
 
-    assert sim_config["time"]["end"]["year"] == 2020
-    assert sim_config["input_data"]["artifact_path"] == "~/vivarium.yaml"
+    initial_job_params = job_parameters.branch_configuration.copy()
+    update_dict = parameter_update_format(job_parameters)
+    sim_config = setup_sim(job_parameters).configuration.to_dict()
+    job_config = job_parameters.branch_configuration
+
+    def compare_dicts(test_dict, ref_dict):
```
rmudambi, line 39: This is basically checking that every key in the `ref_dict` exists in the `test_dict` and is matches the `ref_dict` value? I think it'd be good to add a comment to state that explicitly, because I originally read this and expected it to be checking the that the dicts are equal, which this is not doing.

rmudambi (approved):

PR #192, opened 2023-10-05
Prioritize jobs by filling up scenarios, then seeds, then draws

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: [MIC-4569](https://jira.ihme.washington.edu/browse/MIC-4569)

### Changes and notes
 
This change allows psimulate runs to approximately prioritize jobs in the following order:
for a draw, seed pair, fill out all branch configurations
then for a draw fill out every seed
then fill out draws

It turns out that this is already the way that the jobs are ordered in the keyspace, so all we need to do is allocate them across queues round-robin rather than chunk the jobs into subsequences.
### Testing

Added a test that the job list is in input_draw, random_seed, branch_config order
added a test that jobs are allocated to queues in correct order


tests/psimulate/redis_dbs/test_registry.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+
+from vivarium_cluster_tools.psimulate.redis_dbs.registry import RegistryManager
+
+
+@pytest.mark.parametrize("num_queues", [1, 5, 7, 10, 25, 50])
+@pytest.mark.parametrize("num_jobs", [1, 10, 1000, 2005, 10000, 20000])
```
rmudambi, line 7: How long does this test take to run? Do we really need such large numbers? If this is because of having 1000 jobs per queue, why not mock out that number and replace it with something smaller? That actually seems safer anyway, because the numbers you've chosen for test cases are based on that and would potentially not be as interesting if the DEFAULT_JOBS_PER_QUEUE were to change.

tests/psimulate/redis_dbs/test_registry.py
```diff
@@ -0,0 +1,25 @@
...
+    # mock the QueueManager class
+    mocker.patch("vivarium_cluster_tools.psimulate.redis_dbs.registry.QueueManager")
+    queues = [(f"queue_{i}", i) for i in range(num_queues)]
+    manager = RegistryManager(queues, 0)
+    # create a list of dummy jobs based on a list of 100 integers
+    jobs = [{"job": i} for i in range(num_jobs)]
+    # make a numpy array of all the jobs in every queue
+    jobs_by_queue = list(manager.allocate_jobs(jobs))
+    test_jobs = []
+    # Simulate the order that a single worker would pick up jobs round-robin
```
rmudambi, line 18: I'm having a hard time understanding what is happening here. Isn't what we care about that the `ith` element of each queue is less than the `i+1th` element of every queue for all `i`?

tests/psimulate/test_branches.py
```diff
@@ -52,3 +55,20 @@ def test_calculate_random_seeds_existing(seed_count):
...
+
+
+def test_keyspace_order():
+    ## divide an integer range of 15 into five bins
+    input_draw, random_seed, foo, bar, baz = [list(range(i, i + 3)) for i in range(0, 15, 3)]
+    branches = [
+        {"foo": foo, "bar": bar, "baz": baz} for foo, bar, baz in product(foo, bar, baz)
+    ]
+    # Define the correct ordering of the keyspace
+    ordered_keys = product(input_draw, random_seed, branches)
```
rmudambi, line 67: It's not clear to me what this ordering actually is. I believe you when you say it's what we want it to be, but is there a way to make this clearer?

rmudambi (commented):
The solution looks good, but I have some comments regarding the tests.

tests/psimulate/redis_dbs/test_registry.py
patricktnast: It takes about 2s when I run it. I didn't specifically choose the numbers based on the 1000 jobs per queue thing--#185 removed that and replaced it with a number of workers per queue, so queues now could have an arbitrary number of jobs. I just chose numbers that seemed like plausible real-world values. It probably doesn't need to include the 20000, i just chose the largest number I could that still took a few seconds to run.

patricktnast: removed some large values and added some smaller "pathological" ones

tests/psimulate/redis_dbs/test_registry.py
patricktnast: per conversation, I made the test do this instead

tests/psimulate/test_branches.py
patricktnast: I removed 'bar' and 'baz' and substituted a list comprehension for `product`--then, we don't have write out the list in full, but the ordering is more explicit based on the order of iterables.

tests/psimulate/redis_dbs/test_registry.py
```diff
@@ -0,0 +1,21 @@
+import pytest
+
+from vivarium_cluster_tools.psimulate.redis_dbs.registry import RegistryManager
+
+
+@pytest.mark.parametrize("num_queues", [1, 7, 10, 19, 37])
+@pytest.mark.parametrize("num_jobs", [1, 10, 42, 1337, 2023])
```
mattkappel, line 7: elite

tests/psimulate/test_branches.py
```diff
@@ -1,7 +1,10 @@
+from itertools import product
```
mattkappel, line 1: Doesn't look like you're using this anymore, right?

mattkappel (approved):

PR #202, opened 2024-01-04
Adjust Queue Logging

patricktnast:
### Description

- *Category*: 
bugfix / documentation
- *JIRA issue*: [MIC-MIC-4650](https://jira.ihme.washington.edu/browse/MIC-MIC-4650)

### Changes and notes
 
From the ticket:

> Overview: The psimulate logger previously has buckets for pending,  running, and workers that come from when we submitted all jobs to slurm at once. Now, the meaning is somewhat different, or at least ambiguous, because "pending" means all queued jobs, not necessarily the set of jobs (or rather, in this case, workers) which have been submitted to slurm but are waiting to receive resources. 

As I imply above, thinking about the "jobs" submitted to the cluster is a sort of category error--it's the _workers_ that are really getting submitted. Ultimately, i thought the most sensible thing would be to leave "pending" as-is, under the understanding that this now means just "idle jobs awating workers". Instead, at the "Queue All" level, I added a listing for "inactive workers", that is, the number of "missing workers" from the number that were intially submitted, that are either a) waiting for cluster resources or b) have completed all jobs in the queue and quit themselves.

I don't think there's a lazy way to disambiguate (a) from (b), you'd need to pull information from outside of the registry--and I figured the two situations can be resolved by context (one happens mostly at the beginning of the sim, and one happens mostly at the end).

>Also, "running" and "workers" should effectively be the same, unless rq.Worker.all also counts workers in the queue that are not performing a job (but in that case they should soon quit?)

I too out "workers", assuming the same information is given by "running"

>Also consider renaming "Finished" to "Successful" or "Completed"

I renamed this job status (but not the underlying FinishedQueue) to Successful for the purposes of logging.

I also changed some log info statements to debug which IMO deserve it
### Testing

Tested against nutrition optimization, but I am not actually getting allocated any workers atm

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
+++ b/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
@@ -40,7 +40,7 @@ class QueueManager:
             "pending": 0,
             "running": 0,
             "failed": 0,
-            "finished": 0,
+            "successful": 0,
             "done": 0.0,
             "workers": 0,
         }
@@ -98,7 +98,7 @@ class QueueManager:
         self._update_status()
         template = (
             f"Queue {self.name} - Total jobs: {{total}}, % Done: {{done:.2f}}% "
-            f"Pending: {{pending}}, Running: {{running}}, Failed: {{failed}}, Finished: {{finished}} "
+            f"Pending: {{pending}}, Running: {{running}}, Failed: {{failed}}, Successful: {{successful}} "
             f"Workers: {{workers}}."
         )
         if not (self.completed or self.failed):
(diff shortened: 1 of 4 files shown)


rmudambi (commented):
Can you include a screenshot of what the logging output looks like when you are able to get scheduled? I'll approve after I see it look as expected.

N.B. - you can use `psimulate test sleep` rather than running an actual simulation to reduce our cluster impact.

stevebachmeier:
Like Rajan, I'd also like to actually see a screenshot of the logging in action.

src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
```diff
@@ -81,15 +81,15 @@ def enqueue(self, jobs: List[dict], workhorse_import_path: str) -> None:
...
+        self._logger.debug(f"Checking queue {self.name}")
         finished_jobs = self._get_finished_jobs()
         start = time.time()
         results = []
         for job_id in finished_jobs:
             result = self._get_result(job_id)
             if result is not None:
                 results.append(result)
-        self._logger.info(
+        self._logger.debug(
```
stevebachmeier, line 92: I like this change

src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
```diff
@@ -136,21 +135,23 @@ def _update_status(self) -> None:
...
                 self._status["total"] = (
                     len(pending_ids)
                     + len(running_ids)
                     + len(failed_ids)
-                    + self._status["finished"]
+                    + self._status["successful"]
                 )
                 self._status["workers"] = q_workers
-                self._status["done"] = 100 * self._status["finished"] / self._status["total"]
+                self._status["done"] = (
```
stevebachmeier, line 148: so "done" for this context is basically finished + successful?

patricktnast:
psim-log1
psim-log-with-queues
@rmudambi @stevebachmeier here is a screenshot of the logs (with and without individual queues)

src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
patricktnast: Yeah, I don't think that the "done" percentage is including failed jobs

stevebachmeier (approved):

rmudambi (approved):
Looks good. To be clear, we expect these two constraints to hold?
- `max_workers == inactive_workers + running`
- `total_jobs == pending + running + failed + successful`

PR #207, opened 2024-01-10
central logging

hussain-jafari:
### Description
- *Category*: feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-4751

### Changes and notes
Append performance logs and runner data to central logs.  

### Testing
Used a breakpoint to check that performance logs and runner data were being recorded as expected (using schema for reference). To test appending, created small files and copied and pasted exact logic and checked that data was correct size, files were generated correctly, and first file with data was recorded correctly when 1) all data fit in most recent file, 2) no data fit in most recent file, 3) some data fit in most recent file and one new file had to be created, and 4) some data fit in most recent file and multiple new files had to be created.


src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+    )  # only one job number
+    runner_data["project_name"] = output_directory.parents[6].stem
+    runner_data["root_path"] = output_directory.parents[3]
+    runner_data["original_run_date"] = output_directory.parents[2].stem
+    full_run_date = output_directory.parents[0].stem
+    runner_data["run_date"] = full_run_date[: full_run_date.rindex("_")]
+    runner_data["run_type"] = full_run_date[full_run_date.rindex("_") + 1 :]
+    runner_data["log_summary_file_path"] = first_file_with_data
+    runner_data["original_log_file_path"] = (output_directory / "log_summary.csv").as_posix()
+    runner_data_file = Path(performance_logs_dir) / "runner_data.csv"
```
hussain-jafari, line 270: What do we think of this name?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -210,6 +293,8 @@ def report_performance(
     # Set index to include branch configuration/scenario columns
     perf_df, scenario_cols = set_index_scenario_cols(perf_df)
 
+    append_perf_data_to_central_logs(perf_df, output_directory)
```
hussain-jafari, line 296: I think we'll want a try-except block actually. Where would we want to put this? Here, or within the function itself?

stevebachmeier: if you want to try the entire function, I'd wrap it here as opposed to internally and then having to indent everything.

(and I agree we should try/except this thing at least until we get a handle on how stable it is. The question then is what do we do if this portion fails? Do we go back and try and add the failed logs to the main log or just fix the issue and let it go?)

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
     logger.info(f"\n{report_df}")
 
 
+def append_perf_data_to_central_logs(perf_df: pd.DataFrame, output_directory: Path) -> None:
+    """Append performance data to the central logs."""
+    perf_df = perf_df.reset_index()
```
stevebachmeier, line 196: Do you need to make a copy of `perf_df` or is your intention to also append this stuff to the non-central logs?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+    # add location data to perf_df
+    artifact_path_col = "scenario_input_data_artifact_path"
+    if artifact_path_col in perf_df.columns:  # if we parallelized across artifact paths
+        perf_df["location"] = perf_df[artifact_path_col].apply(
+            lambda filepath: Path(filepath).stem
+        )
+    else:  # else get from output directory
+        perf_df["location"] = output_directory.parents[3].stem
```
stevebachmeier, line 204: Won't this usually look like the model spec name, e.g. `nih_us_cvd`? In that case, is "location" the correct col name?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+    unique_scenario_cols = [
+        col for col in all_scenario_cols if not col.startswith("scenario_run_configuration")
+    ]
+    perf_df["scenario_parameters"] = perf_df[unique_scenario_cols].to_dict(orient="records")
+    perf_df["scenario_parameters"] = perf_df["scenario_parameters"].apply(json.dumps)
+    perf_df = perf_df.drop(all_scenario_cols, axis=1)
+
+    # append child job data
+    NUM_ROWS_PER_FILE = 100_000
+    performance_logs_dir = Path("/mnt/team/simulation_science/pub/performance_logs/")
```
stevebachmeier, line 218: I'd add this string to `/psimulate/paths.py`. That's rpobably not quite the correct place for it either but better than a string floating around I think.

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+    log_files = glob.glob(Path(performance_logs_dir).as_posix() + "/log_summary_*.csv")
+    sorted_files = sorted(log_files, key=lambda x: Path(x).stem.replace("log_summary_", ""))
```
stevebachmeier, line 220: Why do you need to ignore "log_summary_" at all? Aren't all the files like log_summary_0001.csv, log_summary_0002.csv, etc so sort should work as-is?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+    most_recent_file_path = sorted_files[-1]
+    most_recent_data = pd.read_csv(most_recent_file_path)
+
+    data_fits_in_file = (len(most_recent_data) + len(perf_df)) <= NUM_ROWS_PER_FILE
+
+    if data_fits_in_file:
+        first_file_with_data = most_recent_file_path
```
stevebachmeier, line 227: You don't actually use this here

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+        perf_df.to_csv(most_recent_file_path, mode="a", header=False, index=False)
+    else:
+        # fill up the most recent file (possibly with 0 rows)
+        first_file_with_data = most_recent_file_path
```
stevebachmeier, line 231: You're not using this either

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+        if rows_to_append == 0:
+            formatted_new_index = "{:04}".format(new_index)
+            first_file_with_data = (
+                Path(performance_logs_dir) / f"log_summary_{formatted_new_index}.csv"
+            )
+
+        num_appends = math.ceil(len(perf_df) / NUM_ROWS_PER_FILE)
+
+        for append_num in range(num_appends):
+            formatted_new_index = "{:04}".format(new_index)
```
stevebachmeier, line 251: nit: I prefer `str(new_index).zfill(4)`

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+        new_index = most_recent_index + 1
+
+        # record first file with data as new file if no data was appended to most recent file
+        if rows_to_append == 0:
+            formatted_new_index = "{:04}".format(new_index)
+            first_file_with_data = (
+                Path(performance_logs_dir) / f"log_summary_{formatted_new_index}.csv"
+            )
+
+        num_appends = math.ceil(len(perf_df) / NUM_ROWS_PER_FILE)
```
stevebachmeier, line 248: nit: it's not really the number of appends since it's more like the number of new files with a single append at the start. maybe something like `num_chunks`?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+        first_file_with_data = most_recent_file_path
+        perf_df.to_csv(most_recent_file_path, mode="a", header=False, index=False)
+    else:
+        # fill up the most recent file (possibly with 0 rows)
+        first_file_with_data = most_recent_file_path
+        rows_to_append = NUM_ROWS_PER_FILE - len(most_recent_data)
+        perf_df[:rows_to_append].to_csv(
+            most_recent_file_path, mode="a", header=False, index=False
+        )
+        perf_df = perf_df[rows_to_append:]
```
stevebachmeier, line 236: nit: add a comment here that you are explicitly removing the rows that you've appended to the log file from the running dataframe

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -189,6 +191,86 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
+            first_file_with_data = (
+                Path(performance_logs_dir) / f"log_summary_{formatted_new_index}.csv"
+            )
+
+        num_appends = math.ceil(len(perf_df) / NUM_ROWS_PER_FILE)
+
+        for append_num in range(num_appends):
+            formatted_new_index = "{:04}".format(new_index)
+            new_file = Path(performance_logs_dir) / f"log_summary_{formatted_new_index}.csv"
+            start_idx = NUM_ROWS_PER_FILE * append_num
```
stevebachmeier, line 253: this doesn't work if you've removed some of the first rows by appending to an existing log file. I think you need to reset the index above at line 236, e.g. `perf_data = perf_data[rows_to_append:].reset_index(drop=True)`

src/vivarium_cluster_tools/vipin/perf_report.py
stevebachmeier: This is what you were previously calling index.csv?

stevebachmeier (changes requested):
Aside from nits, everything looks great as long as you confirm you want to be addign all of this stuff to the original (non-central) log files as well.

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: Yeah, it's more like it's storing information about the runner jobs.

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: I should make a copy. I was misunderstanding scope/namespaces and thought this wouldn't modify the dataframe outside of the function, but it was sloppy to use the same name even if that were the case. Will update!

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: I didn't like that either and was looking for a file to put it in but I guess I didn't look very closely haha. Sounds good to me

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: Yeah actually Python should know how to sort the strings

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: Ooo nice I didn't know about this

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: I guess I was thinking of each file as one append. I'll change it to `num_new_files`

src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -14,6 +14,9 @@
 from vivarium_cluster_tools.psimulate import COMMANDS
 
 DEFAULT_LOAD_TESTS_DIR = "/mnt/team/simulation_science/priv/engineering/load_tests"
+central_performance_logs_directory = Path(
```
hussain-jafari, line 17: Caps?

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -130,7 +134,7 @@ def add_squid_api_data(perf_df: pd.DataFrame):
             right_on=["squid_api_job_id"],
         )
     except Exception as e:
-        print(f"Squid API request failed with: {e}")
+        logger.warning(f"Squid API request failed with: {e}")
```
hussain-jafari, line 137: Not really in the purview of this PR but figured it was a small enough change.

src/vivarium_cluster_tools/psimulate/paths.py
stevebachmeier: Should be, yes

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -193,71 +195,84 @@ def print_stat_report(perf_df: pd.DataFrame, scenario_cols: list):
...
-        perf_df["location"] = perf_df[artifact_path_col].apply(
+    if (
+        artifact_path_col in central_perf_df.columns
+    ):  # if we parallelized across artifact paths
+        central_perf_df["location"] = central_perf_df[artifact_path_col].apply(
             lambda filepath: Path(filepath).stem
         )
     else:  # else get from output directory
-        perf_df["location"] = output_directory.parents[3].stem
+        central_perf_df["location"] = output_directory.parents[3].stem
```
stevebachmeier, line 210: Oh nice...I didn't know you could pass in the number of parents. I always went `xxx.parent.parent.parent.parent...`

stevebachmeier (approved):

src/vivarium_cluster_tools/vipin/perf_report.py
```diff
@@ -210,6 +308,11 @@ def report_performance(
     # Set index to include branch configuration/scenario columns
     perf_df, scenario_cols = set_index_scenario_cols(perf_df)
 
+    try:
+        append_perf_data_to_central_logs(perf_df, output_directory)
```
patricktnast, line 312: What is the intended behavior with respect to running `vipin` as a standalone CLI? 

It looks to me like if you run `vipin` manually, it will still try to append the results to the central logging, unless I'm missing something. In that case, I think that the output directory may not necessarily conform to the directory structure you're relying on above to fill runner information.

patricktnast: (IMO I think it would make sense to not do the central logging for manual `vipins`, again assuming you haven't already excluded them somehow)

src/vivarium_cluster_tools/vipin/perf_report.py
hussain-jafari: I actually learned about parents from reading one of your PRs haha

src/vivarium_cluster_tools/vipin/perf_report.py
patricktnast: If you want to keep the CLI `vipin` from logging to the central directory, you could probably move the call `append_perf_data_to_central_logs` to the inside of `try_run_vipin` in `runner.py` (ln 108ish)

You could make `report_performance` return the performance df, then pass it into your new function that way, instead of calling it inside report_performance. Then psimulate would use it, but `vipin` wouldn't. 

Alternatively, if you want to let vipin be able to do this logging you could make it a CLI argument, and use the input directory instead of the output_directory--since `psimulate` is using the same `log_directory` for both of these, you should be able to just use that instead in `append_perf_data_to_central_logs` 

Would either of those work?

hussain-jafari: I like the first idea. I think it could also give us more direct access to some of the info we need in the runner file. PR updates pending

hussain-jafari: I've moved the append logic to psimulate/runner.py the way you suggested.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
         )
 
 
+def append_perf_data_to_central_logs(perf_df: pd.DataFrame, log_path: Path) -> None:
+    """Append performance data to the central logs."""
+    NUM_ROWS_PER_FILE = 100_000
+
+    if not fnmatch(log_path, "/mnt/team/simulation_science/pub/models/*/results/*"):
```
hussain-jafari, line 113: I've added so we can be more confident that we're pulling the info from the output directory as expected. Thoughts?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+        logger.warning(
+            f"Log path {log_path} not in central results directory. Skipping appending central performance logs."
+        )
+        return
+
+    central_perf_df = perf_df.reset_index().copy()
+    # add location data to central_perf_df
+    artifact_path_col = "scenario_input_data_artifact_path"
```
hussain-jafari, line 121: Check where the 'scenario' in this string comes from

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+    if (
+        artifact_path_col in central_perf_df.columns
+    ):  # if we parallelized across artifact paths
+        central_perf_df["location"] = central_perf_df[artifact_path_col].apply(
```
hussain-jafari, line 125: change to artifact name

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+    log_files = glob.glob(
+        CENTRAL_PERFORMANCE_LOGS_DIRECTORY.as_posix() + "/log_summary_*.csv"
+    )
+    most_recent_file_path = sorted(log_files)[-1]
+    most_recent_data = pd.read_csv(most_recent_file_path)
+    first_file_with_data = most_recent_file_path
+
+    data_fits_in_file = (len(most_recent_data) + len(central_perf_df)) <= NUM_ROWS_PER_FILE
+
+    if data_fits_in_file:
```
hussain-jafari, line 157: if case is unnecessary

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+    else:
+        # fill up the most recent file (possibly with 0 rows)
+        rows_to_append = NUM_ROWS_PER_FILE - len(most_recent_data)
+        central_perf_df[:rows_to_append].to_csv(
+            most_recent_file_path, mode="a", header=False, index=False
+        )
+        # remove rows appended to most recent file from dataframe
+        central_perf_df = central_perf_df[rows_to_append:]
+
+        most_recent_index = int(Path(most_recent_file_path).stem.replace("log_summary_", ""))
```
hussain-jafari, line 168: better variable name

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+    )
+    central_perf_df = central_perf_df.drop(all_scenario_cols, axis=1)
+
+    # append child job data
+    log_files = glob.glob(
+        CENTRAL_PERFORMANCE_LOGS_DIRECTORY.as_posix() + "/log_summary_*.csv"
+    )
+    most_recent_file_path = sorted(log_files)[-1]
+    most_recent_data = pd.read_csv(most_recent_file_path)
+    first_file_with_data = most_recent_file_path
```
hussain-jafari, line 153: get right value at first pass

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +106,116 @@ def report_initial_status(
...
+        # record first file with data as new file if no data was appended to most recent file
+        if rows_to_append == 0:
+            formatted_new_index = str(new_index).zfill(4)
+            first_file_with_data = (
+                CENTRAL_PERFORMANCE_LOGS_DIRECTORY / f"log_summary_{formatted_new_index}.csv"
+            )
+
+        num_new_files = math.ceil(len(central_perf_df) / NUM_ROWS_PER_FILE)
+
+        for file_num in range(num_new_files):
```
hussain-jafari, line 180: use while loop

tests/psimulate/data/artifactless_perf_df.csv
```diff
@@ -0,0 +1,13 @@
+host,job_number,task_number,draw,seed,scenario_parameter_one,scenario_run_configuration_run_key_parameter_one,event_start,event_simulant_initialization_start,event_simulation_start,event_results_start,event_end,exec_time_setup_minutes,exec_time_simulant_initialization_minutes,exec_time_main_loop_minutes,exec_time_step_mean_seconds,exec_time_results_minutes,exec_time_total_minutes,counters_cpu_ctx_switches,counters_cpu_interrupts,counters_cpu_soft_interrupts,counters_cpu_syscalls,counters_freq_current,counters_freq_min,counters_freq_max,counters_disk_read_count,counters_disk_write_count,counters_disk_read_bytes,counters_disk_write_bytes,counters_disk_read_time,counters_disk_write_time,counters_disk_read_merged_count,counters_disk_write_merged_count,counters_disk_busy_time,counters_net_bytes_sent,counters_net_bytes_recv,counters_net_packets_sent,counters_net_packets_recv,counters_net_errin,counters_net_errout,counters_net_dropin,counters_net_dropout,counters_time
```
hussain-jafari, line 1: I copied these over from actual summary logs and kept most of the data values. Would it be cleaner to just make these 0s or NaNs (starting at event_start)?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +107,136 @@ def report_initial_status(
...
+        central_perf_df = central_perf_df[NUM_ROWS_PER_CENTRAL_LOG_FILE:]
+        new_file_index += 1
+
+    return first_file_with_data
+
+
+def generate_runner_job_data(
+    central_perf_df: pd.DataFrame, log_path: Path, first_file_with_data: str
+) -> pd.DataFrame:
+    runner_data = pd.DataFrame(
```
rmudambi, line 196: Docstring

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +107,136 @@ def report_initial_status(
...
+        central_perf_df[:NUM_ROWS_PER_CENTRAL_LOG_FILE].to_csv(new_file, index=False)
+        # remove written rows and update index
+        central_perf_df = central_perf_df[NUM_ROWS_PER_CENTRAL_LOG_FILE:]
+        new_file_index += 1
+
+    return first_file_with_data
+
+
+def generate_runner_job_data(
+    central_perf_df: pd.DataFrame, log_path: Path, first_file_with_data: str
```
rmudambi, line 194: Only need job-id here not whole df

rmudambi (changes requested):

src/vivarium_cluster_tools/psimulate/performance_logger.py
```diff
@@ -0,0 +1,188 @@
...
+    child_job_performance_data[:rows_to_append].to_csv(
+        most_recent_file_path, mode="a", header=False, index=False
+    )
+    # remove rows appended to most recent file from dataframe
+    child_job_performance_data = child_job_performance_data[rows_to_append:]
+
+    most_recent_file_index = int(Path(most_recent_file_path).stem.replace("log_summary_", ""))
+    new_file_index = most_recent_file_index + 1
+
+    while len(child_job_performance_data) != 0:
```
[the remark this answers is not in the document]
hussain-jafari: I'm not sure how to write the logic for appending the child performance data with 

case 1 - filling up the rest of an already existing file with a variable number of rows and 
case 2 - filling up new files with a constant number of rows 

into one case within the while loop.

tests/psimulate/test_appending_perf_logs.py
```diff
@@ -0,0 +1,205 @@
...
+    filepath = Path(__file__).parent / "data/artifact_perf_df.csv"
+    df = pd.read_csv(filepath)
+    index_cols = ["host", "job_number", "task_number", "draw", "seed"]
+    scenario_cols = [col for col in df.columns if col.startswith("scenario")]
+    df = df.set_index(index_cols + scenario_cols)
+    return df
+
+
+@pytest.fixture
+def artifactless_perf_df():
```
patricktnast, line 31: you might be able to simplify a little bit with something like
```
@pytest.fixture
def df_name():
    return "artifact_perf_df"

@pytest.fixture
def perf_df(df_name):
    filepath = Path(__file__).parent / f"data/{df_name}.csv"
    df = pd.read_csv(filepath)
    index_cols = ["host", "job_number", "task_number", "draw", "seed"]
    scenario_cols = [col for col in df.columns if col.startswith("scenario")]
    df = df.set_index(index_cols + scenario_cols)
    return df
```
and then maybe not need to do request.getfixturevalue(df_name) when you parameterize later?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -101,14 +105,20 @@ def report_initial_status(
...
-def try_run_vipin(log_path: Path) -> None:
+def try_run_vipin(output_paths: OutputPaths) -> None:
+    log_path = output_paths.worker_logging_root
     try:
-        report_performance(
+        perf_df = report_performance(
             input_directory=log_path, output_directory=log_path, output_hdf=False, verbose=1
         )
     except Exception as e:
         logger.warning(f"Performance reporting failed with: {e}")
```
patricktnast, line 116: maybe add a `return` here? if creating `perf_df` fails, then certainly appending to the central logs would fail too

patricktnast (approved):
A couple structural comments, but overall nothing I'd consider blocking.

rmudambi (approved):
Looks good just missing a test to  find the correct file to append to if there are already multiple log files present.

PR #214, opened 2024-06-11
update psimulate for new results processing

stevebachmeier:
### Description

- *Category*:  refactor, feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5096

### Changes and notes
 
This updates vct/psimulate to work with the new results processing approach.
Changes include:

- Updates to writing out since we now have a dictionary of datasets instead of a single dataset
- Without a single results dataset output.hdf, I instead write out a new 
  finished_sim_metadata.csv that has all of the non-results columns which
  are used by vct to determine job status
- Fixes to the long-standing job status update bug when restarting.

### Testing

pytests pass. I also ran test sims (child model but with only
mortality and base population):

- `psimulate run`
- `psimulate restart` (I ctrl+c'ed out of a sim to test this)
- `psimulate expand`
- --no-batch option
- batching (the batch size is currently hard-coded as 200). The test sim I was
  running was size 8 so I changed batch size to 2.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/model_specification.py
+++ b/src/vivarium_cluster_tools/psimulate/model_specification.py
@@ -11,7 +11,7 @@ from typing import Optional
 
 import yaml
 from layered_config_tree import LayeredConfigTree
-from layered_config_tree.exceptions import ConfigurationError
+from layered_config_tree.exceptions import ConfigurationError, ConfigurationKeyError
 from loguru import logger
 from vivarium.framework.artifact import parse_artifact_path_config
 from vivarium.framework.configuration import build_model_specification
--- a/src/vivarium_cluster_tools/psimulate/results/processing.py
+++ b/src/vivarium_cluster_tools/psimulate/results/processing.py
@@ -58,9 +58,8 @@ def _concat_metadata(
     # Skips all the pandas index checking because columns are in the same order.
     start = time.time()
 
-    to_concat = [df.reset_index(drop=True) for df in new_metadata]
-    if not old_metadata.empty:
-        to_concat += [old_metadata.reset_index(drop=True)]
+    to_concat_old = [old_metadata.reset_index(drop=True)] if not old_metadata.empty else []
+    to_concat = to_concat_old + [df.reset_index(drop=True) for df in new_metadata]
 
     updated = _concat_preserve_types(to_concat)
 
@@ -77,15 +76,15 @@ def _concat_results(
     results = {}
     metrics = {key for new in new_results for key in new.keys()}
     for metric in metrics:
-        to_concat = [
+        to_concat_old = (
+            [old_results[metric].reset_index(drop=True)] if metric in old_results else []
+        )
+        to_concat = to_concat_old + [
             df.reset_index(drop=True)
             for result in new_results
             for met, df in result.items()
             if met == metric
         ]
-
-        if metric in old_results:
-            to_concat += [old_results[metric].reset_index(drop=True)]
 
         results[metric] = _concat_preserve_types(to_concat)
 
--- a/src/vivarium_cluster_tools/psimulate/results/processing.py
+++ b/src/vivarium_cluster_tools/psimulate/results/processing.py
@@ -6,70 +6,122 @@ Results Processing
 Tools for processing and writing results.
 
 """
+
 import time
 from pathlib import Path
-from typing import List, Tuple, Union
+from typing import Dict, List, Tuple
 
 import numpy as np
 import pandas as pd
 from loguru import logger
 
 from vivarium_cluster_tools import utilities as vct_utils
+from vivarium_cluster_tools.psimulate.paths import OutputPaths
 
 
 def write_results_batch(
-    output_directory: Path,
-    written_results: pd.DataFrame,
-    unwritten_results: List[pd.DataFrame],
+    output_paths: OutputPaths,
+    existing_metadata: pd.DataFrame,
+    existing_results: Dict[str, pd.DataFrame],
+    unwritten_metadata: List[pd.DataFrame],
+    unwritten_results: List[Dict[str, pd.DataFrame]],
     batch_size: int = 50,
-) -> Tuple[pd.DataFrame, List[pd.DataFrame]]:
+) -> Tuple[
+    pd.DataFrame, List[pd.DataFrame], Dict[str, pd.DataFrame], List[Dict[str, pd.DataFrame]]
+]:
+    """Write batch of results and finished simulation metadata to disk."""
+    logger.info(f"Writing batch of {batch_size} results.")
+    new_metadata_to_write, unwritten_metadata = (
+        unwritten_metadata[:batch_size],
+        unwritten_metadata[batch_size:],
+    )
     new_results_to_write, unwritten_results = (
         unwritten_results[:batch_size],
         unwritten_results[batch_size:],
     )
-    results_to_write = _concat_results(written_results, new_results_to_write)
+    metadata_to_write = _concat_metadata(existing_metadata, new_metadata_to_write)
+    results_to_write = _concat_results(existing_results, new_results_to_write)
 
     start = time.time()
-    _safe_write_results(results_to_write, output_directory / "output.hdf")
+    # write out updated metadata and results
+    for metric, df in results_to_write.items():
+        _safe_write(df, output_paths.results_dir / f"{metric}.parquet")
+    _safe_write(metadata_to_write, output_paths.finished_sim_metadata)
     end = time.time()
-    logger.info(f"Updated output.hdf in {end - start:.4f}s.")
-    return results_to_write, unwritten_results
+    logger.info(f"Updated results in {end - start:.4f}s.")
+    return metadata_to_write, unwritten_metadata, results_to_write, unwritten_results
 
 
-def _concat_results(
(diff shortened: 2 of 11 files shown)


src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -66,37 +84,55 @@ def process_job_results(
...
             )
             logger.info(f"Unwritten results: {len(unwritten_results)}")
             logger.info(f"Elapsed time: {(time() - start_time) / 60:.1f} minutes.")
-        return status
+
+    return status
 
 
-def load_existing_outputs(result_path: Path, restart: bool) -> pd.DataFrame:
+def load_existing_output_metadata(result_path: Path, restart: bool) -> pd.DataFrame:
```
rmudambi, line 106: Nit:
```suggestion
def load_existing_output_metadata(metadata_path: Path, restart: bool) -> pd.DataFrame:
```

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -66,37 +84,55 @@ def process_job_results(
...
     assert (
-        existing_outputs.empty or restart
+        existing_output_metadata.empty or restart
     ), "How do you have existing outputs on an initial run?"
-    return existing_outputs
+    return existing_output_metadata
+
+
+def load_existing_results(result_path: Path, restart: bool) -> Dict[str, pd.DataFrame]:
+    filepaths = list(result_path.glob("*.parquet"))
```
rmudambi, line 118: Does this have to be converted to a list?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -66,37 +84,55 @@ def process_job_results(
...
     if num_jobs_completed:
         logger.info(
-            f"{num_jobs_completed} of {len(keyspace)} jobs completed in previous run."
+            f"{num_jobs_completed} of {total_num_jobs} jobs completed in previous run."
         )
-    if num_jobs_completed != len(existing_outputs):
-        extra_jobs_completed = num_jobs_completed - len(existing_outputs)
+    if num_jobs_completed != len(finished_sim_metadata):
+        extra_jobs_completed = num_jobs_completed - len(finished_sim_metadata)
         logger.warning(
```
rmudambi, line 136: Out of scope, but are we sure this shouldn't be an error rather than a warning?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -167,6 +203,7 @@ def main(
     # a cartesian product representation of the parameter space and
     # branches is a flat representation with the product expanded out.
     keyspace.persist(output_paths.keyspace, output_paths.branches)
+    total_num_jobs = len(keyspace)
```
rmudambi, line 206: Nit: it's generally preferred to define variables as close as possible to their first usage unless there is some reason not to do so.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -80,20 +90,23 @@ def work_horse(job_parameters: dict) -> pd.DataFrame:
...
+
         idx = pd.MultiIndex.from_tuples(
             [(job_parameters.input_draw, job_parameters.random_seed)],
             names=["input_draw_number", "random_seed"],
         )
-        output_metrics = pd.DataFrame(metrics, index=idx)
-        for k, v in collapse_nested_dict(job_parameters.branch_configuration):
-            output_metrics[k] = v
-        return output_metrics
+        finished_results_metadata = pd.DataFrame(index=idx)
```
rmudambi, line 104: Why are you putting the "input_draw_number" and "random_seed" in the index here, when in `processing._concat_metadata` you drop the index?

rmudambi (commented):
Doesn't need to be in this PR if we have more work planned in VCT, but we are not consistent in using "outputs" versus "results".

src/vivarium_cluster_tools/psimulate/model_specification.py
```diff
@@ -11,7 +11,7 @@
 
 import yaml
 from layered_config_tree import LayeredConfigTree
-from layered_config_tree.exceptions import ConfigurationError
+from layered_config_tree.exceptions import ConfigurationError, ConfigurationKeyError
```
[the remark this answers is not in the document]
stevebachmeier: Yes, on line 72

src/vivarium_cluster_tools/psimulate/results/processing.py
```diff
@@ -6,70 +6,122 @@
...
 
+
+def _concat_metadata(old: pd.DataFrame, new: List[pd.DataFrame]) -> pd.DataFrame:
+    # Skips all the pandas index checking because columns are in the same order.
     start = time.time()
-    _safe_write_results(results_to_write, output_directory / "output.hdf")
+
+    to_concat = [df.reset_index(drop=True) for df in new]
+    if not old.empty:
+        to_concat += [old.reset_index(drop=True)]
```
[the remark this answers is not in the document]
stevebachmeier: Yeah I noticed that. It's what we were previously doing and I don't feel strongly.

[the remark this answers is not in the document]
stevebachmeier: Will do this

src/vivarium_cluster_tools/psimulate/runner.py
stevebachmeier: Create a jira ticket to think about this

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
stevebachmeier: Will look into this

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -169,7 +182,7 @@ def parameter_update_format(job_parameters: JobParameters) -> dict:
         },
         "randomness": {
             "random_seed": job_parameters.random_seed,
-            "additional_seed": job_parameters.input_draw,
+            "additional_seed": job_parameters.input_draw,  # <- FIXME: is this a typo?
```
stevebachmeier, line 185: @rmudambi I didn't change this but it seemed fishy that we're setting additional seed to input_draw here.

I also think this is the thing that still has us having multiple columns of the same seed and draw in the new metadata file:
```
>>> [c for c in metadata2.columns if "draw" in c or "seed" in c]
['input_data.input_draw_number', 'run_configuration.run_key.input_draw', 'run_configuration.run_key.random_seed', 'randomness.random_seed', 'randomness.additional_seed']
```

src/vivarium_cluster_tools/psimulate/results/processing.py
```diff
@@ -52,76 +51,70 @@ def write_results_batch(
...
-        slice_df = pd.DataFrame(data=np.concatenate(slices), columns=columns)
-        for col, dtype in original_dtypes.items():
-            slice_df[col] = slice_df[col].astype(dtype)
-        splits.append(slice_df)
-    return pd.concat(splits, axis=1)[col_order]
+    """Concatenate datasets and preserve all ``numpy`` dtypes (but not any
+    pandas-specific dtypes, e.g. categories become objects). We assume that all
+    dataframes in the list have identical columns and dtypes to the first.
+    """
+    dtype_mapping = {col: df_list[0][col].dtype for col in df_list[0].columns}
```
rmudambi, line 100: What if df_list has no elements? Is this possible?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -97,15 +97,11 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
 
         results = sim.get_results()  # Dict[measure, results dataframe]
 
-        idx = pd.MultiIndex.from_tuples(
-            [(job_parameters.input_draw, job_parameters.random_seed)],
-            names=["input_draw_number", "random_seed"],
-        )
-        finished_results_metadata = pd.DataFrame(index=idx)
+        finished_results_metadata = pd.DataFrame(index=[0])
```
rmudambi, line 100: Is the `index` argument necessary?

rmudambi (approved):

src/vivarium_cluster_tools/psimulate/results/processing.py
stevebachmeier: Nope, it won't ever enter `write_results_batch` unless there are unwritten results and to to_concat will have to have something.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
stevebachmeier: It is! That's the bug I introduced yesterday. Without it, no actual values get added

```
>>> import pandas as pd
>>> df1 = pd.DataFrame()
>>> df1["foo"] = "bar"
>>> df1
Empty DataFrame
Columns: [foo]
Index: []
>>> df2 = pd.DataFrame(index=[0])
>>> df2["foo"] = "bar"
>>> df2
   foo
0  bar
>>> 
```

PR #217, opened 2024-06-28
Bugfix/sbachmei/mic 5139 export pythonhashseed

stevebachmeier:
### Description

- *Category*:  bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5139

### Changes and notes
 It turns out that we don't actually need to export the var
when launching off psimulate sruns - they're passed automatically.
So all I've done is raise early on if it's not set.

### Testing

tests pass. I also tried launching a psimulate w/o it set (all jobs failed)
and w/ it set (...all jobs also failed but NOT from the PYTHONHASHSEED
error and the jobs did indeed start running)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,8 @@ jobs:
         shell: bash -le {0}
     steps:
       - uses: actions/checkout@v3
+      - name: Set PYTHONHASHSEED
+        run: echo "PYTHONHASHSEED=42" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


albrja (approved):

hussain-jafari (approved):

stevebachmeier:
NOTE: We decided not to merge this change because we feel the onus should be on the designer/implementer to ensure they are working with different data structures correctly. We also intend to add e2e tests to test against reproducibility.

PR #221, opened 2024-08-01
Add Simulation Backups to PSimulate

patricktnast:
### Description

- *Category*: 
feature
- *JIRA issue*: [MIC-5203](https://jira.ihme.washington.edu/browse/MIC-5203)

### Changes and notes
 
Add pickled simulation backups to Psimulate.
You can specify the pickling frequency and whether to create backups; backups generally live in a "sim_backups" directory under the results root. The backups are deleted once the simulation completes successfully.
### Testing

wrote some unit tests and did testing with MNCH Child to ensure results are reproducible, and backups are saved, restarted, and deleted as expected.

I'm not quite finished testing the deletion but I think it can still get a review in the meantime.

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/jobs.py
+++ b/src/vivarium_cluster_tools/psimulate/jobs.py
@@ -23,6 +23,7 @@ class JobParameters(NamedTuple):
     input_draw: int
     random_seed: int
     results_path: str
+    backup_configuration: dict
     extras: dict
 
     @property
@@ -31,6 +32,7 @@ class JobParameters(NamedTuple):
         return {
             "model_specification": self.model_specification,
             "results_path": self.results_path,
+            "backup_configuration": self.backup_configuration,
         }
 
     @property
@@ -62,6 +64,8 @@ def build_job_list(
     finished_sim_metadata: pd.DataFrame,
     make_backups: bool,
     backup_freq: int,
+    backup_dir: Path,
+    backup_metadata_path: Path,
     extras: dict,
 ) -> Tuple[List[dict], int]:
     jobs = []
@@ -75,7 +79,13 @@ def build_job_list(
                 input_draw=int(input_draw),
                 random_seed=int(random_seed),
                 results_path=str(output_root),
-                extras={"backup_freq": backup_freq * 60} if make_backups else {},
+                backup_configuration={
+                    "make_backups": make_backups,
+                    "backup_dir": backup_dir,
+                    "backup_freq": backup_freq * 60,
+                    "backup_metadata_path": backup_metadata_path,
+                },
+                extras={},
             )
 
             if already_complete(parameters, finished_sim_metadata):
--- a/src/vivarium_cluster_tools/psimulate/jobs.py
+++ b/src/vivarium_cluster_tools/psimulate/jobs.py
@@ -6,7 +6,7 @@ psimulate Jobs
 """
 
 from pathlib import Path
-from typing import List, NamedTuple, Tuple
+from typing import List, NamedTuple, Optional, Tuple
 
 import numpy as np
 import pandas as pd
@@ -62,8 +62,7 @@ def build_job_list(
     output_root: Path,
     keyspace: branches.Keyspace,
     finished_sim_metadata: pd.DataFrame,
-    make_backups: bool,
-    backup_freq: int,
+    backup_freq: Optional[int],
     backup_dir: Path,
     backup_metadata_path: Path,
     extras: dict,
@@ -80,9 +79,10 @@ def build_job_list(
                 random_seed=int(random_seed),
                 results_path=str(output_root),
                 backup_configuration={
-                    "make_backups": make_backups,
                     "backup_dir": backup_dir,
-                    "backup_freq": backup_freq * 60,
+                    "backup_freq": backup_freq * 60
+                    if backup_freq is not None
+                    else backup_freq,
                     "backup_metadata_path": backup_metadata_path,
                 },
                 extras={},
--- a/src/vivarium_cluster_tools/psimulate/jobs.py
+++ b/src/vivarium_cluster_tools/psimulate/jobs.py
@@ -60,6 +60,8 @@ def build_job_list(
     output_root: Path,
     keyspace: branches.Keyspace,
     finished_sim_metadata: pd.DataFrame,
+    make_backups: bool,
+    backup_freq: int,
     extras: dict,
 ) -> Tuple[List[dict], int]:
     jobs = []
@@ -73,7 +75,7 @@ def build_job_list(
                 input_draw=int(input_draw),
                 random_seed=int(random_seed),
                 results_path=str(output_root),
-                extras={},
+                extras={"backup_freq": backup_freq * 60} if make_backups else {},
             )
 
             if already_complete(parameters, finished_sim_metadata):
--- a/src/vivarium_cluster_tools/psimulate/jobs.py
+++ b/src/vivarium_cluster_tools/psimulate/jobs.py
@@ -80,7 +80,9 @@ def build_job_list(
                 results_path=str(output_root),
                 backup_configuration={
                     "backup_dir": backup_dir,
-                    "backup_freq": backup_freq,
+                    "backup_freq": backup_freq * 60
+                    if backup_freq is not None
+                    else backup_freq,
                     "backup_metadata_path": backup_metadata_path,
                 },
                 extras={},
--- a/src/vivarium_cluster_tools/psimulate/jobs.py
+++ b/src/vivarium_cluster_tools/psimulate/jobs.py
@@ -80,9 +80,7 @@ def build_job_list(
                 results_path=str(output_root),
                 backup_configuration={
                     "backup_dir": backup_dir,
-                    "backup_freq": backup_freq * 60
-                    if backup_freq is not None
-                    else backup_freq,
+                    "backup_freq": backup_freq,
                     "backup_metadata_path": backup_metadata_path,
                 },
                 extras={},
--- a/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
+++ b/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
@@ -10,6 +10,7 @@ Unified interface to multiple Redis Databases.
 import random
 import time
 from collections import defaultdict
+from itertools import chain
 from typing import Any, Dict, Iterator, List, Tuple, Union
 
 import redis
@@ -301,8 +302,6 @@ class RegistryManager:
         return status
 
     def get_params_by_job(self):
-        from itertools import chain
-
         jobs = list(chain.from_iterable([q._queue.jobs for q in self._queues]))
         return {job.id: job.kwargs["job_parameters"] for job in jobs}
 
--- a/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
+++ b/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
@@ -301,7 +301,7 @@ class RegistryManager:
         self._logger.info(template.format(**status))
         return status
 
-    def get_params_by_job(self):
+    def get_params_by_job(self) -> dict[str, dict]:
         jobs = list(chain.from_iterable([q._queue.jobs for q in self._queues]))
         return {job.id: job.kwargs["job_parameters"] for job in jobs}
 
--- a/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
+++ b/src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
@@ -300,5 +300,11 @@ class RegistryManager:
         self._logger.info(template.format(**status))
         return status
 
+    def get_params_by_job(self):
+        from itertools import chain
+
+        jobs = list(chain.from_iterable([q._queue.jobs for q in self._queues]))
+        return {job.id: job.kwargs["job_parameters"] for job in jobs}
+
     def __len__(self) -> int:
         return len(self._queues)
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -18,13 +18,9 @@ with_no_cleanup = click.option(
     help="Hidden developer option, if flagged, don't automatically "
     "cleanup results directory on failure.",
 )
-with_make_backups = click.option(
-    "--make-backups", is_flag=True, help="Periodically save simulation state to disk."
-)
 backup_freq = click.option(
     "--backup-freq",
-    type=int,
-    default=5,
+    default=None,
     show_default=True,
-    help="Interval in hours between saving backups.",
+    help="Interval in seconds between saving backups.",
 )
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -23,7 +23,7 @@ with_no_cleanup = click.option(
 backup_freq = click.option(
     "--backup-freq",
     type=MINUTES_OR_NONE,
-    default=30,
+    default="30",
     show_default=True,
     help="Interval in minutes between saving backups. Set to 'None' or 'none' to disable backups.",
 )
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -20,7 +20,7 @@ with_no_cleanup = click.option(
 )
 backup_freq = click.option(
     "--backup-freq",
-    default=None,
+    default=30,
     show_default=True,
-    help="Interval in seconds between saving backups.",
+    help="Interval in minutes between saving backups.",
 )
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -18,7 +18,7 @@ with_no_cleanup = click.option(
     help="Hidden developer option, if flagged, don't automatically "
     "cleanup results directory on failure.",
 )
-with_sim_backup = click.option(
+with_make_backups = click.option(
     "--make-backups", is_flag=True, help="Periodically save simulation state to disk."
 )
 backup_freq = click.option(
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -8,6 +8,8 @@ Command line options for configuring results handling in psimulate runs.
 """
 import click
 
+from vivarium_cluster_tools.cli_tools import MinutesOrNone
+
 with_no_batch = click.option(
     "--no-batch", is_flag=True, help="Don't batch results, write them as they come in."
 )
@@ -20,7 +22,8 @@ with_no_cleanup = click.option(
 )
 backup_freq = click.option(
     "--backup-freq",
+    type=MinutesOrNone(),
     default=30,
     show_default=True,
-    help="Interval in minutes between saving backups.",
+    help="Interval in minutes between saving backups. Set to 'None' or 'none' to disable backups.",
 )
--- a/src/vivarium_cluster_tools/psimulate/results/cli_options.py
+++ b/src/vivarium_cluster_tools/psimulate/results/cli_options.py
@@ -18,3 +18,13 @@ with_no_cleanup = click.option(
     help="Hidden developer option, if flagged, don't automatically "
     "cleanup results directory on failure.",
 )
+with_sim_backup = click.option("--make_backups",
+                               is_flag=True,
+                               help="Periodically save simulation state to disk.")
+backup_freq = click.option(
+    "--backup_freq",
+    type=int,
+    default=5,
+    show_default=True,
+    help="Interval in hours between saving backups.",
+)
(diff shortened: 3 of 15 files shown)


src/vivarium_cluster_tools/psimulate/results/cli_options.py
```diff
@@ -18,3 +18,13 @@
     help="Hidden developer option, if flagged, don't automatically "
     "cleanup results directory on failure.",
 )
+with_sim_backup = click.option(
```
stevebachmeier, line 21: nit: rename to `make_backups` for consistency w/ other options

src/vivarium_cluster_tools/psimulate/results/cli_options.py
```diff
@@ -18,3 +18,13 @@
...
+    "--make-backups", is_flag=True, help="Periodically save simulation state to disk."
+)
+backup_freq = click.option(
+    "--backup-freq",
+    type=int,
+    default=5,
+    show_default=True,
+    help="Interval in hours between saving backups.",
```
stevebachmeier, line 29: Maybe add to the help string that it is only applicable if `--make-backups`.

Actually...do you need both? Could you just get by w/ `--backup-freq` (where setting to zero would be the default and effectively like the current default of `--make-backups` being false?

src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
```diff
@@ -300,5 +301,9 @@ def update_and_report(self) -> Dict[str, Union[int, float]]:
         self._logger.info(template.format(**status))
         return status
 
+    def get_params_by_job(self):
```
stevebachmeier, line 304: type hint the return

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -203,3 +223,54 @@ def format_and_record_details(
...
+            if time() >= time_to_save:
+                sim.write_backup(backup_path)
+                time_to_save = time() + backup_freq
+    else:
+        sim.run()
+
+
+def get_backup(job_parameters: JobParameters) -> Optional[SimulationContext]:
+    backup_dir = job_parameters.backup_configuration["backup_dir"]
+    if not backup_dir.exists() or not os.listdir(backup_dir):
```
stevebachmeier, line 245: Why do you need to check both?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -203,3 +223,54 @@ def format_and_record_details(
...
+        # No sim backups to load
+        return None
+    metadata_path = job_parameters.backup_configuration["backup_metadata_path"]
+    try:
+        pickle_metadata = pd.read_csv(metadata_path)
+    except FileNotFoundError:
+        return None
```
stevebachmeier, line 252: Maybe a comment here. So this would mean that the sim _is_ creating backups and it's just that this specific job doesn't have one?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -203,3 +223,54 @@ def format_and_record_details(
...
+    query_conditions = f"input_draw == {job_parameters.input_draw} & random_seed == {job_parameters.random_seed}"
+
+    # Add branch parameter conditions to the query
+    for k, v in collapse_nested_dict(job_parameters.branch_configuration):
+        query_conditions += f' & `{k}` == "{v}"'
+
+    # Use the query method to find rows that match the lookup parameters
+    run_ids = pickle_metadata.query(query_conditions)["job_id"].to_list()
+    if run_ids:
+        filename = Path(backup_dir / run_ids[0]).with_suffix(".pkl")
```
stevebachmeier, line 262: There should only ever be one row, right? Is there a way this could potentially be untrue? Might be worth raising if `len(run_ids) > 1`

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -47,36 +50,52 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
...
+            sim = backup
+            event["simulation_start"] = time()
+            exec_time = {
+                "setup_minutes": (event["simulation_start"] - event["start"]) / 60
+            }  # execution event
+            logger.info(
+                f'Simulation setup from backup completed in {exec_time["setup_minutes"]:.3f} minutes.'
+            )
+            logger.info(f"Starting main simulation loop")
+        else:
```
stevebachmeier, line 68: Everything in this else block is the same as before, correct?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -47,36 +50,52 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
...
-            event["simulation_start"] - event["simulant_initialization_start"]
-        ) / 60
-        logger.info(
-            f'Simulant initialization completed in {exec_time["simulant_initialization_minutes"]:.3f} minutes.'
-        )
-
-        logger.info(f"Starting main simulation loop with {num_steps} time steps")
-        sim.run()
+        backup = get_backup(job_parameters)
+        if backup:
```
stevebachmeier, line 57: Is there guaranteed to never be a backup on an initial run? (I think so b/c psimulate always creates a new and unique timestamped results directory.) Is there any situation where a backup might exist but we _don't_ want to use it?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -47,36 +50,52 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
...
+        else:
+            sim = setup_sim(job_parameters)
+
+            start_time = pd.Timestamp(**sim.configuration.time.start.to_dict())
+            end_time = pd.Timestamp(**sim.configuration.time.end.to_dict())
+            step_size = pd.Timedelta(days=sim.configuration.time.step_size)
+            num_steps = int(math.ceil((end_time - start_time) / step_size))
+
+            start_snapshot = CounterSnapshot()
+            event = {"start": time()}  # timestamps of application events
```
stevebachmeier, line 77: delete this; it was already defined in line 54

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -47,36 +50,52 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
...
+            logger.info(f"Restarting simulation from saved backup")
+            sim = backup
+            event["simulation_start"] = time()
+            exec_time = {
+                "setup_minutes": (event["simulation_start"] - event["start"]) / 60
+            }  # execution event
+            logger.info(
+                f'Simulation setup from backup completed in {exec_time["setup_minutes"]:.3f} minutes.'
+            )
+            logger.info(f"Starting main simulation loop")
```
stevebachmeier, line 67: Would be cool if you could provide more information like number of steps left to run (to be similar to how it says how many steps exist on the first non-backup run. This is useful for estimating total sim time on-the-fly

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -47,36 +50,52 @@ def work_horse(job_parameters: dict) -> Tuple[pd.DataFrame, Dict[str, pd.DataFra
...
+            exec_time = {
+                "setup_minutes": (event["simulant_initialization_start"] - event["start"])
+                / 60
+            }  # execution event
+            logger.info(
+                f'Simulation setup completed in {exec_time["setup_minutes"]:.3f} minutes.'
+            )
+
+            sim.initialize_simulants()
+            event["simulation_start"] = time()
```
stevebachmeier, line 90: It seems like this and everything below is duplicated (or very near duplicated) with the backup section as well - can it be pulled out of the if/else to DRY it up a bit?

tests/psimulate/test_runner.py
```diff
@@ -1,13 +1,34 @@
...
 def test_report_initial_status():
-
     number_existing_jobs = 10
     finished_sim_metadata = pd.DataFrame(index=range(number_existing_jobs))
     report_initial_status(number_existing_jobs, finished_sim_metadata, 100)
     with pytest.raises(RuntimeError, match="There are 1 jobs from the previous run"):
         report_initial_status(number_existing_jobs + 1, finished_sim_metadata, 100)
+
+
+def test_write_backup_metadata(tmp_path):
```
stevebachmeier, line 19: should test backup deletion as well (maybe even just as part of this test)

stevebachmeier (approved):
Looks good! This is exciting

src/vivarium_cluster_tools/psimulate/results/cli_options.py
patricktnast: Yeah, i went back and forth a bit about this because I'd prefer to add fewer flags all things being equal but it also seems good to be more explicit

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: I just looked at doing something akin to:
`def log_event(event_dict: dict[str,int], exec_time: dict[str,int], event_name: str, last_event: str) -> None:
    event_dict[event_name] = time()
    exec_time[f"{last_event}_minutes"] = (
                event_dict[event_name] - event_dict[last_event]
            ) / 60
    logger.info(
                f'{last_event.replace("_", " ").capitalize()} completed in {exec_time[f"{last_event}_minutes"]:.3f} minutes.'
            )`
But there isn't a consistent enough logic between the names of events and the string used in the log message for this to be worth it IMO

tests/psimulate/test_runner.py
patricktnast: Added to test_vivarium_work_horse.py

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: > Is there guaranteed to never be a backup on an initial run? (I think so b/c psimulate always creates a new and unique timestamped results directory.)
That sounds correct to me.

> Is there any situation where a backup might exist but we don't want to use it?
Good question. I don't have a use case in mind but it would be better not to preclude this. We could add a flag for it, but the user is also free to just delete any backups if they don't want them used.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: yes, I've moved a couple things around now but there is no functional change.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: That should be the case. One thing I didn't really consider until now was what happens if we save backups, then restart and also save backups with the restarted run. I think VCT as written will overwrite the backup metadata for the *new* run instead of appending to the file, so there is only one job per branch. 

It seems to me that it would be better to append to the csv and then choose the most recently created matching file, do you agree?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: Yep, either because it failed so early that it didnt save or somebody came along and deleted the backup.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: The backup directory doesn't get created at all if we're not making backups, and `os.listdir()` does a FileNotFoundError if the dir does not exist. I guess you could incorporate that into the try except block used later, but it seemed a bit convoluted to do the AFNP style thing in this case

src/vivarium_cluster_tools/psimulate/redis_dbs/registry.py
patricktnast: I'm never sure how far to go down the line with nested dictionaries outside of the layeredconfigtree context.
Should I do dict[str, dict] or try to go down to dict[str, dict[str, Any] or dict[str, dict[str, Union[str, int, dict]]??

src/vivarium_cluster_tools/psimulate/results/cli_options.py
albrja: I think if we want backups to be the default behavior we only need the frequency flag and no backups would be 0. I'd also prefer less flags and I think you could set a bool trigger for logic with the input from frequency so you shouldn't have to change that much.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -153,13 +153,34 @@ def try_run_vipin(output_paths: OutputPaths) -> None:
         logger.warning(f"Appending performance data to central logs failed with: {e}")
 
 
+def write_backup_metadata(
```
albrja, line 156: Is this what we were talking about that could be really slow say if we have 50k jobs?

albrja: I still think it is fine if we only do this once.

tests/psimulate/worker/test_vivarium_work_horse.py
```diff
@@ -48,3 +50,64 @@ def compare_dicts(test_dict, ref_dict):
...
+    mocker.patch(
+        "vivarium_cluster_tools.psimulate.worker.vivarium_work_horse.get_current_job",
+        return_value=mocker.Mock(id="job_id"),
+    )
+    # touch a file
+    (tmp_path / "job_id.pkl").touch()
+    assert (tmp_path / "job_id.pkl").exists()
+    # remove the file
+    remove_backups(tmp_path)
+    assert not (tmp_path / "job_id.pkl").exists()
```
albrja, line 113: I'm not sure how you would do it but it seems useful to test we are in fact loading from the most previous back up point however that would require you to both fail a job and restart in in a test.

albrja (approved):

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
rmudambi: I'm not totally following why not to put this all inside the try/except below. That seems simpler to me and I don't see the downside. 

Also, shouldn't the directory always exist if the backup metadata file exists?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -258,6 +283,12 @@ def main(
     registry_manager.enqueue(
         jobs=job_parameters, workhorse_import_path=worker.WORK_HORSE_PATHS[command]
     )
+
+    if make_backups:
+        write_backup_metadata(
+            backup_metadata_path=output_paths.backup_metadata_path,
+            parameters_by_job=registry_manager.get_params_by_job(),
```
rmudambi, line 290: Just to be clear, this will record all jobs even if we have more jobs queued than our max jobs (e.g. if we submitted 10k jobs with max jobs of 8k, would they all be recorded here, or only the first 8k).

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
rmudambi: Can it instead replace the relevant row, rather than appending? Appending seems fine as well though.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -203,3 +217,53 @@ def format_and_record_details(
...
+def run(sim: SimulationContext, job_parameters: JobParameters):
+    backup_freq = job_parameters.backup_configuration["backup_freq"]
+    if backup_freq:
+        backup_dir = job_parameters.backup_configuration["backup_dir"]
+        backup_path = (backup_dir / str(get_current_job().id)).with_suffix(".pkl")
+        time_to_save = time() + backup_freq
+        while not sim.past_stop_time():
+            sim.step()
+            if time() >= time_to_save:
+                sim.write_backup(backup_path)
```
rmudambi, line 231: See my comments in the vivarium PR

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: The actual backup files given by the backup metadata file aren't guaranteed to exist, so appending has the advantage of selecting the appropriate backup if, say, we do a restart but the job fails before creating a new backup, or the "most recent" backup gets deleted or corrupted.

src/vivarium_cluster_tools/psimulate/results/cli_options.py
patricktnast: done

src/vivarium_cluster_tools/psimulate/runner.py
patricktnast: yes, it's really just defining the mapping between the branch configuration and the job ID, not making any guarantees about what is being run or what files will exist.

src/vivarium_cluster_tools/psimulate/jobs.py
```diff
@@ -62,8 +62,7 @@ def build_job_list(
     output_root: Path,
     keyspace: branches.Keyspace,
     finished_sim_metadata: pd.DataFrame,
-    make_backups: bool,
-    backup_freq: int,
+    backup_freq: Optional[int],
```
rmudambi, line 65: Doesn't this make backups off by default? I would assume we'd want them on by default

patricktnast: fixed

albrja (approved):

tests/psimulate/worker/test_vivarium_work_horse.py
patricktnast: I added to this test to load from potentially multiple existing backups. For unit tests, I think we don't necessarily need to kill an active job and restart it--we can test it from either end "as though" a job failed and we're picking it back up.

It would be great to have some e2e tests for something like this, but i think that would have to be on Jenkins.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -236,17 +238,32 @@ def get_backup(job_parameters: JobParameters) -> Optional[SimulationContext]:
...
+                logger.warning(
+                    f"Multiple backups found for {job_parameters}. Using the most recent and deleting the rest."
+                )
+                for stale_file in set(existing_pickles) - {last_pickle}:
+                    os.remove(stale_file)
+            with open(last_pickle, "rb") as f:
                 sim = dill.load(f)
+            current_job_id = get_current_job().id
+            logger.info(f"Renaming backup file {last_pickle} to {current_job_id}.pkl")
+            if job_parameters.backup_configuration["backup_freq"] is not None:
```
rmudambi, line 257: When would this ever not be the case? I guess if you ran with backups, and then restarted without them? Is that a realistic use-case?

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
```diff
@@ -236,17 +238,32 @@ def get_backup(job_parameters: JobParameters) -> Optional[SimulationContext]:
...
+                # Sleep to prevent FS latency when loading the pickle
+                sleep(5)
+                os.rename(last_pickle, (backup_dir / str(current_job_id)).with_suffix(".pkl"))
             return sim
-    except (OSError, FileNotFoundError):
+    except (OSError, FileNotFoundError, EOFError):
```
rmudambi, line 262: When does the EOFError happen?

rmudambi (approved):

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: >  if you ran with backups, and then restarted without them
Yes. In particular, if we don't use backups, then we don't make / append to the metadata file for that set of jobs. So if we rename the file, a subsequent run would not find the backup. 

Whether or not it's realistic, it's possible, so we should handle it consistently. The other option, though, is to just always write/append jobs to the backup csv file regardless of whether we're actually planning to save backups or not.

src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py
patricktnast: In testing this, it seems that filesystem latency can cause a situation where we try to open a file that was modified (i think the os.rename()) recently enough that it is interpreted as an empty file. and unpiclking gives an EOFError. This is why I put the 5 second sleep in there, which seems to add enough time to avoid this.

I also wanted to catch situations where the pickle file is corrupted, e.g. if the sim crashes "just" during the write of the pickle, I guess it might make sense to try a few attempts before moving on to just restarting the sim.

PR #255, opened 2025-04-14
Albrja/mic 5234/worker log verbosity

albrja:
### Add click option to make log level for individual simulations configurable
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-5234

### Changes and notes
-adds click option for logging verbosity for individual simulations

### Testing
Manually test different options and default options. Added test for new feature.


src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -82,12 +82,29 @@ def psimulate():
     "configuration file.",
     callback=cli_tools.coerce_to_full_path,
 )
+# TODO: MIC-5234 - use this where other cli cluster options are being used
+@click.option(
+    "--logging-verbosity",
```
albrja, line 87: There is no l flag for location and I didn't want to do something too similar to the normal pdb for the main process.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -82,12 +82,29 @@ def psimulate():
...
+    "-l",
+    type=click.Choice(
+        [
+            "0",
+            "1",
+            "2",
+        ],
+    ),
+    required=False,
+    default="0",
```
albrja, line 97: This is the default of InteractiveContext

src/vivarium_cluster_tools/psimulate/jobs.py
```diff
@@ -96,7 +96,7 @@ def build_job_list(
                     "backup_freq": backup_freq,
                     "backup_metadata_path": backup_metadata_path,
                 },
-                extras={},
+                extras=extras,
```
albrja, line 99: I think this was actually a small bug but we weren't using it anyways.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -82,12 +82,29 @@ def psimulate():
     "configuration file.",
     callback=cli_tools.coerce_to_full_path,
 )
+# TODO: MIC-5234 - use this where other cli cluster options are being used
+@click.option(
```
stevebachmeier, line 86: We'll want this in the other commands as well (expand, restart, etc). Should probably add it to `shared_options` via `cli_tools`

src/vivarium_cluster_tools/psimulate/cli.py
stevebachmeier: What about `-s, --sim-verbosity` and then we could do `-vvv -sss`?

stevebachmeier: Regardless, it's probably worth slighly tweaking the description of the normal verbosity `-vvv` to clarify that it's the verbosity of the runner and to use `sss` (or whatever you end up w/) for simulation-specific verbosity.

src/vivarium_cluster_tools/psimulate/cli.py
albrja: It gets attached to the sim so it should be in those sims since it gets put in the simulation instance.

src/vivarium_cluster_tools/psimulate/cli.py
albrja: I like sim verbosity. I think I would rather have a level as the input versus counting the s or sss like pdb that way it is very clear they are different.

stevebachmeier (approved):

tests/psimulate/worker/test_logging.py
```diff
@@ -0,0 +1,56 @@
...
+        ):
+            super().__init__()
+            self._logging_level = logging_verbosity
+
+    sim = ParallelSimulationContext(
+        job_parameters.model_specification,
+        configuration=job_parameters.sim_config,
+        logging_verbosity=job_parameters.extras["sim_verbosity"],
+    )
+    assert sim._logging_level == log_level
```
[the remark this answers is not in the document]
hussain-jafari: Actually I'm kind of confused by this. Won't this just always pass no matter how logging_verbosity is used in the source code?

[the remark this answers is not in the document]
albrja: The sim verbosity will always get passed. This wasn't configurable before but is an argument for InteractiveContext so this is preserving the default behavior of InteractiveContext but also making it a click options.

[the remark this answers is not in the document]
hussain-jafari: Oh OK I see yeah if we're just testing the click option feature this is fine.

hussain-jafari (approved):

PR #261, opened 2025-06-06
type-hint cli_tools.py

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6082

### DIFF
--- a/src/vivarium_cluster_tools/cli_tools.py
+++ b/src/vivarium_cluster_tools/cli_tools.py
@@ -11,11 +11,14 @@ from typing import TypeVar
 
 import click
 
-R = TypeVar("R")
-Decorator = Callable[[Callable[..., R]], Callable[..., R]]
+# NOTE: The argument type hints for the cli wrappers are not precise; they should
+# be type-hinted using Protocols. However, the functions being wrapped are never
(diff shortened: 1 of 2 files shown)


src/vivarium_cluster_tools/cli_tools.py
```diff
@@ -11,10 +10,10 @@
 
 import click
 
-Decorator = Callable[[Callable], Callable]
+Decorator = Callable[[Callable[..., object]], Callable[..., object]]
 
 
-def with_verbose_and_pdb(func: Callable) -> Callable:
+def with_verbose_and_pdb(func: Callable[..., object]) -> Callable[..., object]:
```
rmudambi, line 16: Doesn't this destroy the type hinting of the wrapped function? Shouldn't we use a TypeVar here to show that the signature of the wrapped function hasn't changed?

stevebachmeier: I guess it does, yeah - this is just the Callable equivalent of using Any (... for inputs and object for the return). I'm still not very comfortable using TypeVars but I'll give it a shot

albrja (approved):

PR #272, opened 2025-08-15
sbachmei-hjafari-rmudambi/review copilot typing

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-XYZ

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -4,12 +4,14 @@ Cluster Interface
 =================
 
 """
+from __future__ import annotations
 
 import atexit
 import os
 import shutil
 from pathlib import Path
-from typing import NamedTuple, TextIO
+from tempfile import _TemporaryFileWrapper
+from typing import NamedTuple
 
 from vivarium_cluster_tools.psimulate.environment import ENV_VARIABLES
 from vivarium_cluster_tools.utilities import get_drmaa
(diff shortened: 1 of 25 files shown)


src/vivarium_cluster_tools/psimulate/cluster/interface.py
```diff
@@ -30,33 +32,26 @@ class NativeSpecification(NamedTuple):
     queue: str
     peak_memory: int  # Memory in GB
     max_runtime: str
-    hardware: list[str | None]
+    hardware: list[str]
```
patricktnast, line 35: It seems we are not reproducing the same behavior in this file, as before we were filtering a mixed list of strings and Nones, and now we're basically just checking for non-empty list. Is there a justification for why we can say that the list will never have None vals?

src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -65,19 +65,16 @@ def from_entry_point_args(
...
 
         """
-        if result_directory is None:
-            raise ValueError("Result directory must be provided.")
 
-        # Handle the result_directory conversion separately since it has different type requirements
-        result_dir_path = (
-            result_directory if isinstance(result_directory, Path) else Path(result_directory)
-        )
+        results_directory_path = cls._coerce_path(result_directory)
```
patricktnast, line 69: I don't think I buy the wisdom of using an internal method to replace an awkward line of code

patricktnast (approved):

src/vivarium_cluster_tools/psimulate/paths.py
patricktnast: oh, whoops, you are right, I misread this as not being part of this class

PR #312, opened 2026-04-09
Albrja/mic 6947/psim config

albrja:
### Support configuration file for psimulate commands
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6947

### Changes and notes
-support configuration file for psimulate commands
-support keyword args for existing positional arguments. Single letter options for existing positional arguments will all be capital.
-Deprecation warnings if positional arguments are used

### Testing
-Ran multiple psimulate run instances with different config files and command line args.


src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -51,15 +51,37 @@ def psimulate() -> None:
 
 
 @psimulate.command()
+@cli_tools.with_run_config
```
albrja, line 54: This looks kind of messing with currently supporting positional and keyword args but that will go away in the future.

src/vivarium_cluster_tools/cli_tools.py
```diff
@@ -94,3 +97,121 @@ def convert(
 
 
 MINUTES_OR_NONE = MinutesOrNone()
+
```
albrja, line 100: Changes in this file are the main purpose of this pr. Loading a new config file and parsing its args plus any command line args.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -106,6 +130,26 @@ def run(
...
+        "--model-specification/-M",
+    )
+    branch_configuration = cli_tools.resolve_deprecated_positional(
+        branch_configuration_pos,
+        branch_configuration_opt,
+        "branch_configuration",
+        "--branch-configuration/-B",
+    )
+
+    if model_specification is None:
```
rmudambi, line 147: Should we not also error if `branch_configuration` is `None`?

tests/psimulate/test_positional_arg_deprecation.py
rmudambi, line 1: We typically group tests into modules that map onto the source code modules that are being tested. The AI likes to group them into tests of features, and that's what I'm seeing here. We should adhere to our existing practice. If we think there's a better organizational structure, we should discuss this as a team to adopt.

rmudambi (approved):
This looks great, but I think you need to reorganize the tests.

hussain-jafari (approved):

PR #313, opened 2026-04-09
Albrja/mic-6946/write-requested-resources

albrja:
### Write out file with requested resources for psimulate commands
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6946

### Changes and notes
-write out file with the user requested resources for psimulate ocmmands

### Testing
Ran psimulate command. Log message and output file were produced.


hussain-jafari (approved):

rmudambi (changes requested):
This isn't what I thought we had agreed on. I thought we agreed we would write this out in the format of the configuration file that you began supporting in https://github.com/ihmeuw/vivarium_cluster_tools/pull/312

tests/psimulate/test_runner.py
```diff
@@ -73,3 +135,222 @@ def test_write_backup_metadata(tmp_path: Path) -> None:
         }
     )
     assert_frame_equal(metadata, expected_df)
+
+
+def test_write_configuration_run_command(tmp_path: Path) -> None:
```
albrja, line 140: There is a lot of test code but it really is not doing that much. There are 5 new tests which are:
-testing the config file works for run, restart, and expand
-it verifies that users can do a mix of CLI options and config file options
-verifies the output configuration file and be reused for the next psimulate command.

A lot of the code is generating run parameters/configuration and verifying values.

albrja:
Output file:
image

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -98,6 +99,90 @@ def write_backup_metadata(
...
+    backup_freq: int | None,
+    extra_args: dict[str, Any],
+) -> None:
+    """Write the resolved run configuration to a YAML file in the output directory.
+
+    This creates a ``configuration.yaml`` file that records all of the
+    parameters used for the run.  The file is written in a format that is
+    directly usable with ``psimulate <command> --run-config configuration.yaml``
+    so that team members can reproduce or review a previous run without
+    needing to know the original command-line arguments.
```
rmudambi: Nit: 
```suggestion
    so that previous runs can be easily reproduced.
```

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -98,6 +99,90 @@ def write_backup_metadata(
...
+
+    # Input paths – keys match the names accepted by --run-config
+    if command == COMMANDS.run:
+        if input_paths.model_specification is not None:
+            config["model_specification"] = str(input_paths.model_specification)
+        if input_paths.branch_configuration is not None:
+            config["branch_configuration"] = str(input_paths.branch_configuration)
+        config["result_directory"] = str(input_paths.result_directory)
+        if input_paths.artifact is not None:
+            config["artifact_path"] = str(input_paths.artifact)
```
rmudambi, line 150: When would these ever be `None`? Aren't model spec and branch configuration required?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -98,6 +99,90 @@ def write_backup_metadata(
...
+        # restart / expand – the result directory *is* the results_root
+        config["results_root"] = str(input_paths.result_directory)
+
+    # Cluster resources
+    config["project"] = native_specification.project
+    config["queue"] = native_specification.queue
+    config["peak_memory"] = native_specification.peak_memory
+    config["max_runtime"] = native_specification.max_runtime
+    if native_specification.hardware:
+        config["hardware"] = ",".join(native_specification.hardware)
```
rmudambi, line 161: Is this right? I don't remember how you parse this key when running a simulation from a config file - I would have expected this to be a yaml list: i.e.
```yaml
hardware:
    - "hardware-value-1"
    - "hardware-value-2"
```

rmudambi (approved):
Some minor comments, but overall looks good.

src/vivarium_cluster_tools/psimulate/runner.py
albrja: They input args can be None because you can provide them as a CLI option or via a config file. This is resolving which one is used - the user provided it via the command and this is overriding the configuration to be written out. There is a separate check that will raise a runtime error if they are None, so yes you are correct but that is not what is happening here.

src/vivarium_cluster_tools/psimulate/runner.py
albrja: The click options is a string that is comma separated by values e.g. "hardware-value1, hardware-value2" and so it would just be a string in a yaml file. I admit this is a little confusing but this is to implemented to work with the click option as currently implemented.

PR #314, opened 2026-04-14
Albrja/mic 6957/config parser

albrja:
### Add necessary classes to parse pipeline configuration yaml
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6957

### Changes and notes
-adds classes to parse pipeline configuration


src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+SUPPORTED_STEP_TYPES = {"pytest", "notebook", "python", "shell"}
+
+REQUIRED_PIPELINE_FIELDS = {"name", "project", "queue", "output_directory", "steps"}
+
+
+@dataclass
+class ResourceConfig:
+    """Compute resource specification for a pipeline step."""
+
+    memory: float | None = None
```
stevebachmeier: What are the units of all of these? docstrings and also consider renaming `memory` to `memory_gb`.

For runtime, do we want to support hh:mm:ss format? Or just hours? If just hours, then maybe call it `runtime_hours`.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+                f"Step '{self.name}': unsupported type '{self.type}'. "
+                f"Must be one of {sorted(SUPPORTED_STEP_TYPES)}."
+            )
+
+        # type requires path
+        if self.type is not None and self.path is None:
+            raise ValueError(f"Step '{self.name}': 'type' requires 'path' to be provided.")
+
+        # Must not have both command and type+path
+        if self.is_raw_command and self.is_structured:
```
stevebachmeier: Would it be cleaner/clearer if instead of this branching pattern we just had two more concrete data classes, e.g. StructuredStepConfig and CommandStepConfig?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+SUPPORTED_STEP_TYPES = {"pytest", "notebook", "python", "shell"}
```
stevebachmeier: Add a comment saying what all these things are

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        """True if the step uses type + path."""
+        return self.type is not None and self.path is not None
+
+    @property
+    def is_raw_command(self) -> bool:
+        """True if the step uses a raw command string."""
+        return self.command is not None
+
+    def _validate(self) -> None:
+        """Validate this step's internal consistency. Raise on errors."""
```
stevebachmeier:
```suggestion
        """Validate this step's internal consistency."""
```

src/vivarium_cluster_tools/psimulate/pipeline_config/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+========================
+Pipeline Configuration
+========================
+
+Configuration and utilities for pipeline workflow orchestration.
+
+"""
+
+from vivarium_cluster_tools.psimulate.pipeline_config.config import (
+    PipelineConfig,
```
stevebachmeier: What about "workflow" instead of "pipeline" throughout? This would distinguish it from vivarium's Pipelines, continue to be generic (i.e. not mentioning "jobmon"), and you also mentioned that the psimulate command would be "workflow" so would be consistent with that

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+            queue=pipeline["queue"],
+            output_directory=Path(pipeline["output_directory"]),
+            default_environment=pipeline.get("default_environment"),
+            steps=steps,
+        )
+        config._validate()
+        return config
+
+    def _validate(self) -> None:
+        """Validate pipeline-level constraints. Raise on errors."""
```
stevebachmeier:
```suggestion
        """Validate pipeline-level constraints."""
```

tests/psimulate/workflow_config/utilities.py
```diff
@@ -0,0 +1,89 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import pytest
+import yaml
+
+
+def make_step_dict(**overrides: Any) -> dict[str, Any]:
```
stevebachmeier, line 11: This kind of stuff doesn't _really_ belong in a conftest. Maybe add to a utilities dir?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+    def is_structured(self) -> bool:
+        """True if the step uses type + path."""
+        return self.type is not None and self.path is not None
+
+    @property
+    def is_raw_command(self) -> bool:
+        """True if the step uses a raw command string."""
+        return self.command is not None
+
+    def _validate(self) -> None:
```
stevebachmeier: Consider running this on `__post_init__` and then you wouldn't need to explicitly run it in the pipeline class

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+
+@dataclass
+class StepConfig:
+    """Configuration for a single pipeline step."""
+
+    name: str
+    command: str | None = None
+    type: str | None = None
```
stevebachmeier: This collides with the more generic `type()` method. Mbye `step_type`? You can still call it `type` in the config and just map it.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        """Load, validate, and return a PipelineConfig from a YAML file."""
+        with open(path) as f:
+            raw = yaml.safe_load(f)
+
+        pipeline = raw["pipeline"]
+
+        # Check required top-level fields
+        for field_name in REQUIRED_PIPELINE_FIELDS:
+            if field_name not in pipeline:
+                raise ValueError(
```
stevebachmeier: KeyError

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        # Check required top-level fields
+        for field_name in REQUIRED_PIPELINE_FIELDS:
+            if field_name not in pipeline:
+                raise ValueError(
+                    f"Pipeline configuration is missing required field '{field_name}'."
+                )
+
+        raw_steps = pipeline["steps"]
+        if not raw_steps:
+            raise ValueError("Pipeline 'steps' must not be empty.")
```
stevebachmeier: KeyError

tests/psimulate/pipeline_config/test_config.py
```diff
@@ -0,0 +1,215 @@
...
+        yaml_path = write_pipeline_yaml(tmp_path, data)
+        config = PipelineConfig.from_yaml(yaml_path)
+        assert config.steps[0].path == "tests/test_one.py"
+
+
+class TestStepConfigProperties:
+    """Verify ``StepConfig`` classification properties."""
+
+    def test_is_structured_true(self) -> None:
+        step = StepConfig(name="s", type="pytest", path="tests/")
```
stevebachmeier: It looks like only the pytest type is tested at all. Should probaby add a parametrized test against all of the supported types (so that it automatically tests everything as we add or remove supported types in the future)

tests/psimulate/pipeline_config/test_config.py
```diff
@@ -0,0 +1,215 @@
...
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import pytest
+
+from tests.psimulate.pipeline_config.conftest import (
+    make_pipeline_dict,
```
stevebachmeier: Various missing tests
1. missing a test for a step without resources
2. No test for a yaml missing the `pipeline` key
3. ResourceConfig
4. Including both `command` and `path`

tests/psimulate/workflow_config/utilities.py
stevebachmeier, line 11: This can be accidentally used to create bad combinations, e.g. `make_step_dict(command="sleep 3")` would make a dict with both `type`/`path` (from the defaults) and a `command`. Maybe split into two maker functions?

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        """Validate this step's internal consistency. Raise on errors."""
+        # Validate step type if provided
+        if self.type is not None and self.type not in SUPPORTED_STEP_TYPES:
+            raise ValueError(
+                f"Step '{self.name}': unsupported type '{self.type}'. "
+                f"Must be one of {sorted(SUPPORTED_STEP_TYPES)}."
+            )
+
+        # type requires path
+        if self.type is not None and self.path is None:
```
stevebachmeier: Need to check the opposite as well. Also, 'args' isn't supported w/ `command`, right?

```
if self.type is None and self.path is not None:
    raise(...)

if self.is_raw_command and (self.path is not None or self.args is not None):
    raise ValueError(
        f"Step '{self.name}': 'path' and 'args' are not used with 'command'."
    )
```

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+    project: str
+    queue: str
+    output_directory: Path
+    default_environment: str | None
+    steps: list[StepConfig]
+
+    @classmethod
+    def from_yaml(cls, path: Path) -> PipelineConfig:
+        """Load, validate, and return a PipelineConfig from a YAML file."""
+        with open(path) as f:
```
stevebachmeier: path is a pathlib.Path object - use `path.open()`

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -0,0 +1,151 @@
...
+        config._validate()
+        return config
+
+    def _validate(self) -> None:
+        """Validate pipeline-level constraints. Raise on errors."""
+        # Unique step names
+        names = [s.name for s in self.steps]
+        if len(names) != len(set(names)):
+            raise ValueError(
+                f"Step names must be unique. Duplicate names found: {set([name for name in names if names.count(name) > 1])}"
```
stevebachmeier: nit: `{name for name in names if ...}`

stevebachmeier (changes requested):
Some nits, but there are some test gaps. Also, need to not ignore `path` or `args` when also passing in `command`

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: Good call. I think hh:mm:ss since there is a good chance some steps may be pretty short.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: IMO it isn't that complicated right now and we could always refactor it later.

tests/psimulate/workflow_config/utilities.py
albrja: I'm not sure I understand what you are proposing for two different helper functions. I also think we test both getting values of each attribute but also these specific configuration errors you mention here so I'm not sure why this is a problem. This is a helper method to make there be less code when generating the tests.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+    """Unique name for this step within the workflow."""
+    resources: ResourceConfig
+    """Resource configuration for this step."""
+    command: str | None = None
+    """Raw command string to execute for this step. Mutually exclusive with 'type' and 'path'."""
+    type: str | None = None
+    """Structured step type (e.g. 'pytest', 'notebook'). Requires 'path' to be provided."""
+    path: str | list[str] | None = None
+    """Path(s) to the module or directory for structured steps. Required if 'type' is provided."""
+    args: str | None = None
```
rmudambi: As discussed, remove these three arguments for now, and we'll add them back as structured types later.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
rmudambi: +1

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+# and on the backend developers can choose how these are implemented, leaving room for future flexibility.
+
+REQUIRED_WORKFLOW_FIELDS = {"name", "project", "queue", "output_directory", "steps"}
+
+
+@dataclass
+class ResourceConfig:
+    """Compute resource specification for a workflow step."""
+
+    memory_gb: float | None = None
```
rmudambi: What happens if `memory_gb` is `None`? Shouldn't it always have to be provided (i.e. not have a default value)?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+    """Memory in GB."""
+    runtime: str | None = None
```
rmudambi: If you don't provide a default value for `runtime`, you won't have to check if it's `None` in the `__post_init__`.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+
+import yaml
+
+SUPPORTED_STEP_TYPES = {"pytest", "notebook", "python", "shell"}
+# NOTE: Each step type will map to a specific execution strategy. Pytest will run pytest
+# test suites, notebook will execute Juypter notebooks, python will run Python scripts,
+# and shell will execute raw shell commands. Users will only need to know the support types,
+# and on the backend developers can choose how these are implemented, leaving room for future flexibility.
+
+REQUIRED_WORKFLOW_FIELDS = {"name", "project", "queue", "output_directory", "steps"}
```
rmudambi, line 19: Shouldn't `resources` be a required workflow field?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+        raw_steps = workflow["steps"]
+        if not raw_steps:
+            raise ValueError("Workflow 'steps' must not be empty.")
+
+        steps = []
+        for step_dict in raw_steps:
+            step_name = step_dict["name"]
+            raw_resources = step_dict.get("resources")
+            if raw_resources is None:
+                raise ValueError(f"Step '{step_name}': 'resources' is required.")
```
rmudambi: If `resources` becomes a required field, we don't need these lines and can just have `resources=ResourceConfig.from_dict(step_dict["resources"])` below.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
rmudambi: +1

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+            project=workflow["project"],
+            queue=workflow["queue"],
+            output_directory=Path(workflow["output_directory"]),
+            default_environment=workflow.get("default_environment"),
+            steps=steps,
+        )
+        config._validate()
+        return config
+
+    def _validate(self) -> None:
```
rmudambi: This could also be in a `__post_init__`

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -0,0 +1,194 @@
...
+        """Validate workflow-level constraints."""
+        # Unique step names
+        names = [step.name for step in self.steps]
```
rmudambi, line 125: Nit: we could also validate that the queue (and maybe the project?) is valid.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This is getting removed for the time being.

rmudambi (changes requested):

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: Those are in the StepConfig. This is the top level of the workflow. 
```
workflow:
queue: "all.q"
project: "proj_simsci"
output_directory: "~/scratch"
steps:
  name: step1
  resources:
      memory_gb: 2
```

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This happens later when we implement the CLI command.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -138,39 +84,29 @@ def from_yaml(cls, path: Path) -> WorkflowConfig:
...
         if not raw_steps:
-            raise ValueError("Workflow 'steps' must not be empty.")
+            raise KeyError("Workflow 'steps' must not be empty.")
 
         steps = []
         for step_dict in raw_steps:
-            step_name = step_dict["name"]
-            raw_resources = step_dict.get("resources")
-            if raw_resources is None:
-                raise ValueError(f"Step '{step_name}': 'resources' is required.")
```
rmudambi, line 163: I think we still want something like this otherwise the user will get an unhelpful error message if they don't provide one to a step.
```python
if not "resources" in step_dict:
    raise ValueError(f"Step '{step_name}': 'resources' is required.")
```

rmudambi (approved):

stevebachmeier (approved):

PR #320, opened 2026-04-16
Albrja/mic 6958/build pipeline

albrja:
### Build jobmon workflow from configuration
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6958

### Changes and notes
-build jobmon workflow from configuration

### Testing
-Generated jobmon workflow and successfully ran workflow via Jobmon runner
image

image


CHANGELOG.rst
```diff
@@ -1,3 +1,9 @@
+**3.2.0 - TBD/TBD/TBD**
```
albrja, line 1: Forgot the epic changelog so ill start it here.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import TYPE_CHECKING, Callable
+
+from jobmon.client.api import Tool
+
+from vivarium_cluster_tools.psimulate.pipeline_config.config import PipelineConfig, StepConfig
+
+if TYPE_CHECKING:
+    from jobmon.client.workflow import Workflow
+
+COMMAND_RESOLVERS: dict[str, Callable[[str | list[str] | None, str | None], str]] = {
```
stevebachmeier: docstring - wth is going on here?

src/vivarium_cluster_tools/psimulate/pipeline_config/builder.py
```diff
@@ -0,0 +1,130 @@
...
+        f"papermill {_get_single_path(path)} {{output_directory}}/executed/"
+        f"{Path(_get_single_path(path)).name} {args or ''}"
+    ).strip(),
+    "python": lambda path, args: f"python {_join_paths(path)} {args or ''}".strip(),
+    "shell": lambda path, args: f"bash {_join_paths(path)} {args or ''}".strip(),
+}
+
+
+class PipelineWorkflowBuilder:
+    """Builds a complete Jobmon workflow from a pipeline configuration.
```
stevebachmeier:
```suggestion
    """Build a complete Jobmon workflow from a pipeline configuration.
```

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+
+    def __init__(self, config: PipelineConfig) -> None:
+        self.config = config
+
+    def build(self) -> Workflow:
+        """Build the full pipeline DAG and return the Jobmon Workflow."""
+        tool = Tool(name="vivarium_cluster_tools")
+
+        task_template = tool.get_task_template(
```
stevebachmeier: This is a bigger thing that is perhaps not in scope of this, but since this is the second place where we're setting up jobmon workflows, we may want to wrap all of the jobmon stuff so that there's only one place to fix/update things as needed.

I'd make a ticket to do this and we can prioritize accordingly

CHANGELOG.rst
```diff
@@ -1,3 +1,9 @@
...
+
+- Psimulate workflow epic
```
stevebachmeier: Don't mention an "epic" in the changelog. Also, this is too much detail. It might be a simple one-liner, e.g. `- Feature: Implement serial workflows via 'psimulate workflow'` or something like that

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
+"""
+========================
```
stevebachmeier: Match section divider lengths to titles (I think this might actually fail doc-building)

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
...
+        f"papermill {_get_single_path(path)} {{output_directory}}/executed/"
+        f"{Path(_get_single_path(path)).name} {args or ''}"
+    ).strip(),
+    "python": lambda path, args: f"python {_join_paths(path)} {args or ''}".strip(),
+    "shell": lambda path, args: f"bash {_join_paths(path)} {args or ''}".strip(),
+}
+
+
+class WorkflowBuilder:
+    """Builds a complete Jobmon workflow from a workflow configuration.
```
stevebachmeier:
```suggestion
    """Build a complete Jobmon workflow from a workflow configuration.
```

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
...
+        task_template = tool.get_task_template(
+            template_name="workflow_command_step",
+            command_template="conda run --no-capture-output -n {env} {command}",
+            node_args=["command"],
+            task_args=[],
+            op_args=["env"],
+            default_cluster_name="slurm",
+        )
+
+        # TODO: MIC-6997
```
stevebachmeier: What is this todo? leave a brief description

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
...
+        workflow = tool.create_workflow(
+            name=self.config.name,
+            default_cluster_name="slurm",
+            default_max_attempts=3,
+        )
+
+        tasks = []
+        for step in self.config.steps:
```
stevebachmeier, line 67: This means that the steps must be in order in the config, correct? That should be documented somewhere

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
...
+                name=step.name,
+                compute_resources=compute_resources,
+                env=env,
+                command=command,
+            )
+            tasks.append(task)
+
+        # Wire sequential dependencies: each step depends on the previous
+        for i in range(1, len(tasks)):
+            tasks[i].add_upstream(tasks[i - 1])
```
stevebachmeier, line 99: this is cool, I don't think jobmon had an `add_upstream` method back when I used it

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,132 @@
...
+
+def resolve_command(step: StepConfig) -> str:
+    """Resolve a step's configuration into a shell command string.
+
+    For raw command steps, returns the command as-is.
+    For structured steps, uses the type to infer the command.
+    """
+    # Check command directly for type narrowing
+    if step.command is not None:
+        return step.command
```
stevebachmeier, line 137: Gap here if `step.command is not None & (step.type is not None | step.path is not None)`

stevebachmeier (changes requested):
I left some of my own comments, but I also recommend running these PRs through the code_reviewer agent as there are a LOT of things it suggests. I'm just going to dump the thing in its entirety here. We need to discuss as a team how to actually handle this sort of thing.

```
PR Review: Build Jobmon workflow from configuration (#320)

Summary
=======
This PR adds a PipelineWorkflowBuilder class that translates a PipelineConfig (from the base branch's config-parser
work) into a Jobmon Workflow with sequentially-chained tasks. It includes a COMMAND_RESOLVERS dispatch table
for structured step types (pytest, notebook, python, shell), a resolve_command() public function, and unit tests
covering the happy paths. The PR also updates __init__.py exports and the changelog.

Design
=====
{output_directory} placeholder in notebook resolver is never wired to Jobmon template args — builder.py L27-29.
The notebook resolver produces {output_directory}/executed/... in the command string, but output_directory is not
declared in node_args, task_args, or op_args (L53-58). If this is meant for Jobmon substitution, it must be declared
and passed at create_task time. If it's meant for something else, its substitution mechanism is undocumented. This
is likely a runtime bug.

Strictly linear DAG is undocumented — builder.py L88-89. All steps are chained sequentially. This is a reasonable
first pass, but the class is generically named and neither the class docstring nor build() docstring mentions this
limitation. If non-linear DAGs are foreseeable, an optional depends_on field on StepConfig would be far cheaper to
add now than to retrofit later. At minimum, document the sequential assumption.

Tool is re-created on every build() call — builder.py L52. Jobmon's Tool constructor can do server-side registration.
If build() is called more than once, this produces duplicate registrations. Consider creating the tool once in __init__
or as a module-level singleton.

Maintainability
===========
Magic numbers for resource defaults — builder.py L79-85. The values 4 (memory), "01:00:00" (runtime), 1 (cores),
and 3 (max_attempts at L64) are inline literals with no named constants. The unit for memory is ambiguous (GB?
MB?). Extract to module constants with documented units, or push defaults into ResourceConfig field defaults.

Unrecognized step.type raises an unhelpful KeyError — builder.py L129. COMMAND_RESOLVERS[step.type] raises a
bare KeyError for typos like "pytst". The ValueError at L131 only triggers when type or path is None. Use .get() and
raise a ValueError listing valid types:

_get_single_path silently drops extra paths — builder.py L103-106. If a notebook step receives path=["a.ipynb",
"b.ipynb"], only the first is used with no warning. Consider raising when len(path) > 1 for the notebook type, or at
minimum documenting this behavior.

Hardcoded "slurm" cluster name in two places — builder.py L57 and L63. Extract to a constant or make configurable
via PipelineConfig.

DRY
===
Three command resolvers share an identical pattern — builder.py L24-33. pytest, python, and shell all follow f
{executable} {_join_paths(path)} {args or ''}".strip(). Extract to a shared helper:

Repeated resources and resources.<field> guard pattern — builder.py L79-84. The null-checking pattern is repeated
three times with an inconsistency: memory and runtime check both [resources](vscode-file://vscode
app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron
browser/workbench/workbench.html) and resources.field (so memory=0 falls back to default), while cores only
checks [resources](vscode-file://vscode
app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron
browser/workbench/workbench.html) (so cores=0 passes through). Push defaults into ResourceConfig field defaults
to eliminate this pattern entirely.

Tests
====
No test for resolve_command error paths — test_builder.py. The ValueError for missing command/type and the
KeyError for unknown types are untested.

No test for environment fallback chain — The 3-level fallback (step.environment → config.default_environment →
$CONDA_DEFAULT_ENV → "base") has four branches; none are explicitly tested. This is the most important
runtime-affecting logic in build().

No test for compute resource defaults — test_builder.py. The three_step_config fixture uses raw-command steps
with no resources. No test asserts that compute_resources passed to create_task contains the expected defaults.

No test for single-step or empty-step pipelines — These are degenerate inputs that exercise the range(1, len(tasks))
loop boundary.

No test for notebook with list path — Silently dropping extra paths should be covered, at minimum to document the
intentional behavior.

Documentation
===========
resolve_command error message references "bespoke name" which doesn't exist — builder.py L131. The message
says "has no command, type, or recognized bespoke name" but no bespoke name concept exists in the code.
Remove the misleading phrase.

resolve_command is a public function missing Parameters, Returns, and Raises sections — Per project conventions,
this should have a full NumPy-style docstring.

PipelineWorkflowBuilder.build() docstring doesn't mention sequential chaining — The most important behavioral
characteristic of this method is undocumented.

_get_single_path docstring is vague — "Extract a single path string from various path formats" doesn't explain that
lists are truncated to the first element.

Functionality
=========
Inconsistent falsy-value handling in resource defaults — builder.py L79-84. resources.memory if resources and
resources.memory else 4 treats memory=0 as falsy and defaults to 4. But resources.cores if resources else 1 would
pass through cores=0. While 0-valued resources are unlikely, the inconsistency suggests the pattern wasn't
intentional.

{output_directory} in notebook commands will produce broken shell commands at runtime — (Overlaps with Design
#1.) The literal {output_directory} will appear in the shell command. Unless Jobmon is configured elsewhere to
substitute this (which the task template definition doesn't support), notebooks will fail with a file-not-found error
pointing at a literal {output_directory} directory.

Minor Nits
========
[CHANGELOG.rst](vscode-file://vscode
app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron
browser/workbench/workbench.html) L4 typo: "Parse configuration to for workflow" — stray "to". Should be "Parse
configuration for workflow".

[CHANGELOG.rst](vscode-file://vscode
app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron
browser/workbench/workbench.html) L5 trailing whitespace: "Build Jobmon workflow from configuration " has a
trailing space.

Overall
=====
The core builder logic is clean and the test structure is good. The highest-priority issues are:

#1/#20: The {output_directory} placeholder in notebook commands appears to be a bug — it's not wired as a
Jobmon template variable and will produce broken shell commands.
#5: Unknown step.type produces a confusing KeyError instead of a helpful message.
#9/#19: Resource default logic has an inconsistency between memory/runtime and cores.
#11-13: Key behavioral logic (environment resolution, resource defaults) lacks test coverage.
```

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
albrja: It was in the docstring of the class but I will update that comment so it is more clear and add a comment in the config module as well.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
albrja: Validation for the StepConfig has happened at this point. So we know at this point it is a command step or a structured step.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,138 @@
...
+            default_max_attempts=3,
+        )
+
+        tasks = []
+        for step in self.config.steps:
+            command = resolve_command(step, self.config.output_directory)
+            env = (
+                step.environment
+                or self.config.default_environment
+                or os.environ.get("CONDA_DEFAULT_ENV", "base")
```
rmudambi: Should we allow running in the base conda environment?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -23,25 +23,25 @@
 class ResourceConfig:
     """Compute resource specification for a workflow step."""
 
-    memory_gb: float | None = None
-    """Memory in GB."""
-    runtime: str | None = None
-    """Maximum runtime in 'hh:mm:ss' format."""
+    memory_gb: float = 4
+    """Memory in GB. Default is 4."""
```
rmudambi: Per my comment in the other PR, should we support a default value. Maybe we should force the user to think about their resource consumption.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -23,25 +23,25 @@
...
             raise ValueError(f"Invalid runtime '{self.runtime}'. Expected format 'hh:mm:ss'.")
 
     @classmethod
     def from_dict(cls, data: dict[str, Any]) -> ResourceConfig:
         """Create a ResourceConfig from a dictionary."""
         return cls(
-            memory_gb=data.get("memory_gb"),
-            runtime=data.get("runtime"),
+            memory_gb=data.get("memory_gb", 4),
+            runtime=data.get("runtime", "01:00:00"),
```
rmudambi: You're specifying the same default twice by doing this - assuming we want to keep defaults, this is something that would need to be kept in sync and is a potential bug surface.

rmudambi (commented):
Not approving, since changes will need to be made following the comments in the previous PR.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: I like supporting a default runtime that is pretty short but I think requiring memory is good idea.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -23,10 +23,10 @@
 class ResourceConfig:
     """Compute resource specification for a workflow step."""
 
-    memory_gb: float = 4
-    """Memory in GB. Default is 4."""
+    memory_gb: int
+    """Memory in GB."""
     runtime: str = "01:00:00"
-    """Maximum runtime in 'hh:mm:ss' format. Default is '01:00:00'."""
+    """Maximum runtime in 'hh:mm:ss' format."""
```
rmudambi: Did you mean to remove the default from the docstring?

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,144 @@
...
+def _join_paths(path: str | list[str] | None) -> str:
+    """Normalize path to a space-separated string."""
+    if path is None:
+        return ""
+    if isinstance(path, list):
+        return " ".join(str(p) for p in path)
+    return str(path)
+
+
+def resolve_command(step: StepConfig, output_directory: Path) -> str:
```
rmudambi: We don't need this method at all anymore, since we are only supporting simple commands right now.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,144 @@
...
+        # TODO: MIC-6997 - encapsulate Jobmon UI in one place
+        workflow = self._tool.create_workflow(
+            name=self.config.name,
+            default_cluster_name="slurm",
+            default_max_attempts=3,
+        )
+
+        tasks = []
+        for step in self.config.steps:
+            command = resolve_command(step, self.config.output_directory)
```
rmudambi: Can remove this.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,144 @@
...
+                "memory": resources.memory_gb,
+                "runtime": resources.runtime,
+                "cores": resources.cores,
+            }
+
+            task = task_template.create_task(
+                name=step.name,
+                compute_resources=compute_resources,
+                env=env,
+                command=command,
```
rmudambi:
```suggestion
                command=step.command,
```

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -0,0 +1,144 @@
...
+# Mapping of types to functions that resolve a command string from the step's configuration
+COMMAND_RESOLVERS: dict[str, Callable[[str | list[str] | None, str | None, Path], str]] = {
+    "pytest": lambda path, args, output_directory: f"pytest {_join_paths(path)} {args or ''}".strip(),
+    "notebook": lambda path, args, output_directory: (
+        f"papermill {_get_single_path(path)} {output_directory}/executed/"
+        f"{Path(_get_single_path(path)).name} {args or ''}"
+    ).strip(),
+    "python": lambda path, args, output_directory: f"python {_join_paths(path)} {args or ''}".strip(),
+    "shell": lambda path, args, output_directory: f"bash {_join_paths(path)} {args or ''}".strip(),
+}
```
rmudambi: Don't need these anymore. The logic for resolving commands will eventually live in the specific `StepConfig` sub-class for each step type.

rmudambi (approved):
Approved assuming you address the comments I made.

stevebachmeier (approved):

PR #321, opened 2026-04-16
Albrja/mic 6959/workflow subcommand

albrja:
### Rollout of workflow subcommand
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6959

### Changes and notes
-add workflow subcommand to psimulate group
-small updates to configuration to keep cli argument behavior consistent across psimulate commands

### Testing
-ran a workflow with "psimulate workflow -c pipeline_config.yaml"
image
image

Output configuration of requested resources exists in correct format:
image


src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -17,7 +17,7 @@
 
 SUPPORTED_STEP_TYPES = {"pytest", "notebook", "python", "shell"}
 
-REQUIRED_PIPELINE_FIELDS = {"name", "project", "queue", "output_directory", "steps"}
+REQUIRED_PIPELINE_FIELDS = {"name", "steps"}
```
albrja, line 20: This is to keep the workflow subcommand work the same as other psimulate commands for the same args. For example, Ouput directory has to be provided by either the config file or as a cli options. These were required before since the subcommand did not exists but I wanted to have all arguments have validation.

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -39,6 +39,17 @@ def from_dict(cls, data: dict[str, Any] | None) -> ResourceConfig | None:
             cores=data.get("cores", 1),
         )
 
+    def to_dict(self) -> dict[str, Any]:
```
albrja, line 42: This makes it so we can just write this to the output requested resources yaml and not have to parse the config twice.

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -85,15 +96,41 @@ def _validate(self) -> None:
         if not self.is_raw_command and not self.is_structured:
             raise ValueError(f"Step '{self.name}': must provide 'command' or 'type'+'path'.")
 
+    def to_dict(self) -> dict[str, Any]:
```
albrja, line 99: Same here

src/vivarium_cluster_tools/psimulate/pipeline_config/config.py
```diff
@@ -149,3 +188,25 @@ def _validate(self) -> None:
             raise ValueError(
                 f"Step names must be unique. Duplicate names found: {set([name for name in names if names.count(name) > 1])}"
             )
+
```
albrja, line 191: Same here

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -412,3 +412,69 @@ def test(
             "num_workers": num_workers,
         },
     )
+
+
+@psimulate.command()
+@cli_tools.with_pipeline_config
```
albrja, line 419: When the positional arguments for the other psimulate commands are dropped and not just deprecated, all commands will have a similar config decorator like this - which allows users to specify required args in either a config file or from the ommand line.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -35,10 +35,74 @@
...
+        input_paths=None,
+        native_specification=cluster.NativeSpecification(
+            job_name=pipeline_config.name,
+            project=pipeline_config.project,
+            queue=pipeline_config.queue,
+            peak_memory=4,  # Default, steps have their own resources
+            max_runtime="01:00:00",
+            hardware=[],
+        ),
+        max_workers=None,  # Not used for workflow command
```
albrja, line 75: Workflows will be serial steps so this argument is not needed and I do not want users to overwrite it. We hardcode it so only 1 step can run at a time.

src/vivarium_cluster_tools/psimulate/pipeline_config/builder.py
```diff
@@ -58,6 +58,7 @@ def build(self) -> Workflow:
             name=self.config.name,
             default_cluster_name="slurm",
             default_max_attempts=3,
+            max_concurrently_running=1,  # Pipeline steps are sequential
```
albrja, line 61: This would already happen based on how our dag would be defined but I wanted to strictly enforce it.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -412,3 +413,67 @@ def test(
...
+@click.option(
+    "--output-directory",
+    "-o",
+    type=click.Path(file_okay=False),
+    default=None,
+    help="Override output directory from config file.",
+    callback=cli_tools.coerce_to_full_path,
+)
+@cli_tools.with_verbose_and_pdb
+def workflow(
```
stevebachmeier: I think this should be imperative. Maybe `run_workflow`?

tests/psimulate/test_workflow_e2e.py
```diff
@@ -0,0 +1,203 @@
+"""
```
albrja, line 1: This is a very simple e2e test but there is a lot of code because of the setup of the configuration.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -192,3 +231,25 @@ def _validate(self) -> None:
...
+
+    def to_dict(self) -> dict[str, Any]:
+        """Serialize to a dictionary suitable for YAML output."""
+        result: dict[str, Any] = {
+            "name": self.name,
+        }
+
+        if self.project is not None:
+            result["project"] = self.project
+
```
rmudambi, line 171: Nit: your other `to_dict` methods don't have line breaks between the ifs.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -412,3 +413,83 @@ def test(
...
+    specified via the -c/--config option. The config file specifies
+    all workflow steps, compute resources, and execution order.
+
+    Top-level options like project, queue, and output_directory can
+    be provided in the config file and/or overridden from the command line.
+    """
+    logs.configure_main_process_logging_to_terminal(options["verbose"])
+
+    # Parse the workflow configuration
+    workflow_config = WorkflowConfig.from_yaml(config_path)
```
rmudambi, line 465: I'd suggest passing the CLI arguments to your `WorkflowConfig` class method (probably with a new name), so that you can retain project, queue, and output_directory as required workflow fields and validate their existence appropriately in the class.

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -36,9 +36,73 @@
...
+            job_name=workflow_config.name,
+            project=workflow_config.project,
+            queue=workflow_config.queue,
+            peak_memory=4,  # Default, steps have their own resources
+            max_runtime="01:00:00",
+            hardware=[],
+        ),
+        max_workers=None,  # Not used for workflow command
+        max_attempts=3,
+        backup_freq=None,
```
rmudambi, line 77: Why are these hardcoded?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -36,9 +36,73 @@
...
+        logger.info(f"Monitor progress at: {monitoring_url}")
+
+    wf_status = workflow.run()
+
+    if wf_status != "D":
+        logger.warning(
+            f"Workflow finished with status '{wf_status}' (expected 'D' for DONE)."
+        )
+    else:
+        logger.info(f"Workflow completed successfully. Results in {output_root}")
```
rmudambi, line 103: I assume much of this is the same as in `run`. Can it be DRYed?

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -121,52 +185,66 @@ def write_configuration(
...
-        The psimulate sub-command (e.g. ``"run"``, ``"restart"``, ``"expand"``).
+        The psimulate sub-command (e.g. ``"run"``, ``"restart"``, ``"expand"``, ``"workflow"``).
     input_paths
-        The resolved input file paths.
+        The resolved input file paths. None for workflow command.
     native_specification
         The cluster resource specification.
     max_workers
-        Maximum number of concurrent workers.
+        Maximum number of concurrent workers. Not used for workflow command.
```
rmudambi, line 194: Why is this not used by workflow?

src/vivarium_cluster_tools/cli_tools.py
```diff
@@ -189,6 +189,68 @@ def with_run_config(func: CLIFunction) -> CLIFunction:
...
+    workflow configuration files to provide defaults for CLI options like
+    --project, --queue, and --output-directory.
+    """
+    if value is None:
+        return None
+
+    config_path = Path(value).resolve()
+
+    # Parse the workflow YAML
+    with open(config_path) as f:
```
rmudambi, line 207: This looks very duplicative of the parsing of the workflow file in `WorkflowConfig.from_yaml()`. We shouldn't need to do this twice. Can we encapsulate it within a helper method? Do we maybe not need to do it at all, since `psimulate workflow` is guaranteed to have a config file, whereas `psimulate run` is not?

src/vivarium_cluster_tools/psimulate/runner.py
albrja: This was an attempt to keep thing DRY but it required passing fake args that do not get used when we write out the workflow configuration/resources that are used. I will try to clean it up more.

src/vivarium_cluster_tools/cli_tools.py
albrja: I have updated this. This was all to add the decorator but moving validation to all be in the Workflow config class has cleaned this up.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -21,10 +21,12 @@
     from jobmon.client.workflow import Workflow
 
 # Mapping of types to functions that resolve a command string from the step's configuration
-COMMAND_RESOLVERS: dict[str, Callable[[str | list[str] | None, str | None, Path], str]] = {
+COMMAND_RESOLVERS: dict[
```
albrja: These currently won't be used since the structured steps have been currently removed but will be used in a later pr

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -109,22 +165,132 @@ def from_yaml(cls, path: Path) -> WorkflowConfig:
...
+        Raises
+        ------
+        KeyError
+            If ``project``, ``queue``, or ``output_directory`` cannot be resolved
+            from either the YAML file or CLI arguments.
+        """
+        workflow = cls._parse_yaml_file(path)
+        steps = cls._parse_steps(workflow["steps"])
+
+        resolved_project = project or workflow.get("project")
```
albrja, line 211: Resolving command line and config args in here instead of in cli/runner

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -99,6 +180,25 @@ def write_backup_metadata(
     )
 
 
+def write_workflow_configuration(output_root: Path, workflow_config: WorkflowConfig) -> None:
```
albrja, line 183: I decided it was simpler to have a separate workflow and "main" methods to write out the configuration to avoid a bunch if if/else logic and when we include parallel simulations as a step we can leverage both of them.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
rmudambi: Per a comment in the previous PR, these should be removed.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -117,22 +175,108 @@ def from_yaml(cls, path: Path) -> WorkflowConfig:
...
+        resolved_queue = queue or workflow.get("queue")
```
rmudambi, line 212: This docstring claims queue will default to "all.q" if it hasn't been provided, but it appears to error here.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -85,20 +121,37 @@ class WorkflowConfig:
...
+    max_attempts: int = DEFAULT_MAX_ATTEMPTS
+    """Maximum number of Jobmon task attempts. Default is 2."""
 
-    @classmethod
-    def from_yaml(cls, path: Path) -> WorkflowConfig:
-        """Load, validate, and return a WorkflowConfig from a YAML file."""
+    @staticmethod
+    def _parse_yaml_file(
+        path: Path,
+        extra_required_fields: set[str] | None = None,
```
rmudambi: Why do we have this argument? This method is only ever called in `from_yaml_with_cli_overrides()`, and it does not provide any `extra_required_fields`.

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -412,3 +413,66 @@ def test(
...
+    "-m",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Override maximum Jobmon task attempts from config file.",
+)
+@cli_tools.with_verbose_and_pdb
+def workflow(
+    config_path: Path,
+    output_directory: Path | None,
+    **options: Any,
```
rmudambi, line 451: Why is `output_directory` treated differently from the other options like `queue` and `project`?

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -412,3 +413,66 @@ def test(
...
+)
+@click.option(
+    "--max-attempts",
+    "-m",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Override maximum Jobmon task attempts from config file.",
+)
+@cli_tools.with_verbose_and_pdb
+def workflow(
```
rmudambi, line 448: Are we 100% committed to having a new entry-point called `workflow`? Why can't we just add the `"--config"` option to `run` and trigger a "simple" workflow with only a simulation step if it's not provided. 

Either way, leave it as you have it for now, but let's talk about whether we should combine the entry-points when you implement simulation steps.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: It was to allow future flexibility but due to the small refactor we are doing it won't be necessary in the future.

src/vivarium_cluster_tools/psimulate/cli.py
albrja: When a click option uses a callback, it has to be explicitly specified in the function signature. Same thing happens in our other entry points.

src/vivarium_cluster_tools/psimulate/cli.py
albrja: We can talk about it.

rmudambi (approved):

hussain-jafari (approved):

PR #323, opened 2026-04-22
Base jobmon workflow timeout on runner node

stevebachmeier:
### Description

- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6937

### Changes and notes
 

Calculates the timeout time as 2 minutes less than the runner node's remaining
time and passes that into the workflow run call.

### Testing

added unit tests
manually tested `psimulate run` and `psimulate restart`

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -131,6 +131,16 @@ def get_workflow_timeout_seconds() -> int:
             f"Could not determine remaining SLURM time for job {job_id}: {e}"
         ) from e
 
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"squeue failed for SLURM job {job_id} "
+            f"(exit {result.returncode}): {result.stderr.strip()}"
+        )
+
+    stderr_output = result.stderr.strip()
+    if stderr_output:
+        logger.warning(f"squeue stderr for job {job_id}: {stderr_output}")
+
     if not remaining_str:
         raise RuntimeError(
             f"squeue returned no output for SLURM job {job_id}. "
@@ -172,8 +182,9 @@ def _parse_slurm_time(time_str: str) -> int:
     ValueError
         If ``time_str`` does not match a recognized SLURM time format.
     """
-    # Match optional "D-" prefix followed by colon-separated numeric fields.
-    if not re.fullmatch(r"(\d+-)?\d+(:\d+){0,2}", time_str):
+    # When a day prefix is present, squeue always uses D-HH:MM:SS.
+    # Without a day prefix the format is HH:MM:SS, MM:SS, or SS.
+    if not re.fullmatch(r"(\d+-\d+:\d+:\d+|\d+(:\d+){0,2})", time_str):
         raise ValueError(
             f"Unrecognized SLURM time format: '{time_str}'. "
             "Expected D-HH:MM:SS, HH:MM:SS, MM:SS, or SS."
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -146,19 +146,19 @@ def get_runner_node_remaining_seconds() -> int:
             "Cannot determine remaining allocation time."
         )
 
-    remaining_seconds = _parse_slurm_time(remaining_str) - _SLURM_TIMEOUT_BUFFER_SECONDS
-    if remaining_seconds <= 0:
+    remaining_seconds = _parse_slurm_time(remaining_str)
+    workflow_seconds = remaining_seconds - _SLURM_TIMEOUT_BUFFER_SECONDS
+    if workflow_seconds <= 0:
         raise RuntimeError(
-            f"SLURM allocation has {remaining_str} remaining, which is less than "
-            f"the {_SLURM_TIMEOUT_BUFFER_SECONDS}s safety buffer. "
+            f"SLURM allocation has {remaining_str} ({remaining_seconds}s) remaining, "
+            f"which is less than the {_SLURM_TIMEOUT_BUFFER_SECONDS}s safety buffer. "
             "Not enough time to run a workflow."
         )
     logger.info(
-        f"Detected SLURM allocation with {remaining_str} remaining. "
-        f"Setting workflow timeout to {remaining_seconds}s "
-        f"({_SLURM_TIMEOUT_BUFFER_SECONDS}s buffer)."
+        f"Detected SLURM allocation with {remaining_str} ({remaining_seconds}s) remaining. "
+        f"Setting workflow timeout to {workflow_seconds}s ({_SLURM_TIMEOUT_BUFFER_SECONDS}s buffer)."
     )
-    return remaining_seconds
+    return workflow_seconds
 
 
 def _parse_slurm_time(time_str: str) -> int:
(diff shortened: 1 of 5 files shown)


stevebachmeier:
## PR Review: Dynamically determine jobmon workflow timeout from SLURM runner node

### Summary

This PR adds logic to query the remaining time on the SLURM runner node (via `squeue -o %L`) and pass it as `seconds_until_timeout` to the jobmon `workflow.run()` call. A 120-second buffer is subtracted so the workflow can shut down gracefully before SLURM kills the runner node. The implementation adds two functions (`get_runner_node_remaining_seconds` and `_parse_slurm_time`) to the cluster interface module and includes comprehensive unit tests.

### Design

1. **interface.py - `subprocess.run` return code is not checked.** If `squeue` exits non-zero (stale job ID, scheduler issues), stdout may contain an error message like `slurm_load_jobs error: Invalid job id specified`. This passes the empty-string check and reaches the parser, which raises a confusing `ValueError: Unrecognized SLURM time format` instead of a clear error. *(Flagged by design + maintainability)*
   - **Suggestion:** Add `check=True` to the `subprocess.run` call and catch `CalledProcessError` in the existing `except` block, or explicitly check `result.returncode != 0` and include `result.stderr` in the `RuntimeError`.

2. **interface.py - Regex accepts formats `squeue` never produces.** The pattern `(\d+-)?\d+(:\d+){0,2}` accepts `D-SS` and `D-MM:SS` (e.g., `"2-30"` parses as 2 days + 30 seconds). SLURM's `%L` format uses `D-HH:MM:SS` with all three time fields when a day prefix is present.
   - **Suggestion:** Tighten the regex when a day prefix is present, e.g. `(\d+-\d+:\d+:\d+|\d+(:\d+){0,2})`, or document the looser acceptance as intentional.

3. **interface.py - Function name doesn't match semantics.** `get_runner_node_remaining_seconds()` returns remaining time *minus* the buffer, but the name implies the raw remaining time. A caller reading only the name would misunderstand the return value.
   - **Suggestion:** Rename to `get_workflow_timeout_seconds()` to accurately describe what is returned.

### Maintainability

4. **interface.py - Inline comment documents `UNLIMITED` but it's not handled.** The comment says `squeue` can return `"UNLIMITED"`, but it falls through to `_parse_slurm_time` which raises `ValueError` - an exception type not documented in `get_runner_node_remaining_seconds`'s `Raises` section.
   - **Suggestion:** Add an explicit check for `"UNLIMITED"` and raise a clear `RuntimeError` (e.g., "SLURM allocation has no time limit; cannot determine workflow timeout"), or update the comment to note it's unsupported.

5. **test_interface.py - Tests hardcode the buffer value `120`.** If `_SLURM_TIMEOUT_BUFFER_SECONDS` changes, the expected values in tests silently become wrong.
   - **Suggestion:** Import `_SLURM_TIMEOUT_BUFFER_SECONDS` and use it in expected-value expressions: `assert result == 36000 - _SLURM_TIMEOUT_BUFFER_SECONDS`.

### DRY

6. **interface.py vs interface.py - `_runtime_to_seconds` and `_parse_slurm_time` share identical core logic.** Both split on `":"`, branch on part count, and compute `h*3600 + m*60 + s`. `_parse_slurm_time` is a strict superset (adds day prefix and validation). *(Flagged by design + DRY + maintainability)*
   - **Suggestion:** Refactor `_runtime_to_seconds` to delegate to `_parse_slurm_time`:
     ```python
     @staticmethod
     def _runtime_to_seconds(runtime_str: str) -> int:
         return _parse_slurm_time(runtime_str)
     ```
     This gives `_runtime_to_seconds` input validation for free and eliminates the duplicated parser. The corresponding test parameterizations could also be merged.

### Tests

7. **test_interface.py - Missing boundary test at `buffer + 1`.** Tests cover `remaining < buffer` (60s) and `remaining == buffer` (120s), but there's no test at the exact passing boundary (121s - 1 second of workflow time). A regression changing `<= 0` to `< 0` would go undetected.
   - **Suggestion:** Add:
     ```python
     def test_returns_one_second_when_just_above_buffer(self) -> None:
         completed = _make_squeue_result("00:02:01")  # 121s - 120 = 1
         with patch("subprocess.run", return_value=completed):
             assert get_runner_node_remaining_seconds() == 1
     ```

8. **test_interface.py - No test for non-zero squeue exit code.** The test helper always sets `returncode=0`. This is the most likely real-world failure mode and currently has no defined behavior (see finding 1).

9. **test_interface.py - `patch("subprocess.run")` patches the global rather than the module reference.** This works because the source does `subprocess.run(...)`, but it's broader than necessary and could interfere under thread-level parallelism.
   - **Suggestion:** Patch `vivarium_cluster_tools.psimulate.cluster.interface.subprocess.run` instead.

### Documentation

10. **interface.py - Summary line and body describe the wrong value.** The summary says "Return the number of seconds remaining" but the function returns remaining minus buffer. The body says "The result includes a small buffer" which reads as "buffer is added" rather than "buffer is subtracted."
    - **Suggestion:** Summary: `"Return the number of seconds available for the workflow within the current SLURM allocation."` Body: `"A safety buffer is subtracted from the raw remaining time so that the workflow can shut down gracefully before SLURM terminates the runner node."`

11. **interface.py - `_parse_slurm_time` docstring lists formats the regex doesn't exclusively match.** The docstring says `SS, MM:SS, HH:MM:SS, D-HH:MM:SS` but the regex also accepts `D-SS` and `D-MM:SS`. These should match.

12. **CHANGELOG.rst - "Bugfix" vs. enhancement.** If there was no prior bug (just a missing timeout), "Enhancement" or "Feature" would be more accurate than "Bugfix."

### Functionality

13. **interface.py - `ValueError` escapes despite docstring only documenting `RuntimeError`.** When `_parse_slurm_time` raises `ValueError` (for `UNLIMITED` or malformed output), it propagates uncaught from `get_runner_node_remaining_seconds`. The tests confirm this (`test_raises_for_unlimited` expects `ValueError`), but it contradicts the docstring. Callers catching only `RuntimeError` would miss this.
    - **Suggestion:** Either catch `ValueError` in `get_runner_node_remaining_seconds` and wrap it in `RuntimeError`, or document `ValueError` in the `Raises` section.

14. **interface.py - squeue stderr is silently discarded on success.** If squeue writes a warning to stderr but still produces valid stdout, the warning is lost. Low severity, but worth noting since `result.stderr` is available.

### Minor Nits

15. **test_interface.py - `_make_squeue_result` imports `CompletedProcess` inside the function body** rather than at module level where `subprocess` is already imported.

### Overall

Clean, well-scoped PR. The core design is sound and the test coverage is good. The most important items to address are **finding 1** (check squeue return code), **finding 13** (ValueError escaping contradicts documented raises), and **finding 6** (consolidate the two time parsers). The naming and docstring issues (findings 3, 10) would also improve the API for future consumers.

stevebachmeier:
I fixed everything but 13 which didn't seem worth it

src/vivarium_cluster_tools/psimulate/cluster/interface.py
```diff
@@ -81,12 +86,124 @@ def _runtime_to_seconds(runtime_str: str) -> int:
...
-            m, s = parts
-            return int(m) * 60 + int(s)
-        else:
-            return int(parts[0])
+        return _parse_slurm_time(runtime_str)
+
+
+# Buffer in seconds to subtract from the remaining SLURM time so the jobmon
+# workflow shuts down cleanly before SLURM kills the runner node.
+_SLURM_TIMEOUT_BUFFER_SECONDS = 120
```
rmudambi, line 94: Did you just choose this time arbitrarily? Is it based on anything concrete?

rmudambi (approved):

src/vivarium_cluster_tools/psimulate/cluster/interface.py
stevebachmeier: Yeah, totally arbitrary

albrja (approved):

PR #324, opened 2026-04-23
Albrja/mic 6960/simulation task

albrja:
### Add feature to support custom step type "simulation" to run parallel simulations
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6960

### Changes and notes
-refactor so there is a BaseStepConfig ABC class that command steps (CommandStepConfig) and custom type step config classes will subclass.
-adds SimulationStepConfig class to launch parallel simulation for simulation type step type. This class parses configuration flags and infers a raw "psimulate run" command

### Testing
-ran workflow with a simulation step type.

### Code reviewer
PR Review: SimulationStepConfig — type: simulation workflow step support
Summary
This PR adds first-class support for type: simulation steps in the psimulate workflow YAML configuration. A new SimulationStepConfig dataclass (subclassing BaseStepConfig) fans out into parallel Jobmon tasks — one per (draw, seed, branch) combination — reusing the existing job-generation infrastructure from psimulate run. The PR also introduces a task_runner.py CLI entry point for individual Jobmon tasks, extracts a reusable get_task_list() from build_workflow(), and adds per-task parquet result writing. Accompanying test suites cover config parsing/validation, task runner dispatch, and workflow builder DAG construction.

The overall architecture is sound — the class hierarchy (BaseStepConfig → CommandStepConfig / SimulationStepConfig) is clean, the step-type routing via SUPPORTED_STEP_TYPES is extensible, and the reuse of get_task_list() avoids reimplementing the Jobmon task creation logic.

Design
get_tasks() creates timestamped OutputPaths — problematic for resume and multi-step workflows (config.py:521-530). Each call to OutputPaths.from_entry_point_args() generates a new datetime.now() timestamp. If build() is called twice (e.g. resume scenario) or if two simulation steps exist, they get different paths. This also makes SimulationStepConfig a config-data object with filesystem side effects (touch()), which is architecturally surprising. Suggestion: Lift OutputPaths creation into the builder, passing it to get_tasks(), or generate the timestamp once at build time. (Design + Functionality)

CommandStepConfig.get_tasks() silently drops hardware constraint (config.py:370-379). The compute_resources dict omits self.resources.hardware, while SimulationStepConfig passes it through NativeSpecification.to_jobmon_spec(). A user specifying hardware: ["r650"] on a command step would see it silently ignored. Suggestion: Add if self.resources.hardware: compute_resources["constraints"] = "|".join(self.resources.hardware). (Design + Functionality)

Late validation of queue/project being strings (config.py:547-553). The isinstance(self.resources.queue, str) check happens inside get_tasks(), after config parsing and workflow building. If these are required for simulation steps, enforce it at construction time in _validate(). (Design)

O(n²) DAG dependency edges between consecutive fan-out steps (builder.py:70-77). For two consecutive simulation steps each with k tasks, this creates k² add_upstream calls. At typical scales (hundreds to thousands of tasks) this may be acceptable, but worth verifying with Jobmon. A synthetic barrier task would reduce this to O(k). (Design)

from_dict() / _create_from_dict() indirection (config.py:220-280). The two-layer pattern exists solely to centralise _validate_command_type_exclusivity, a three-line static method. Since _parse_steps() already resolves the concrete class, this could be flattened. (Design — low severity)

supported_arguments() abstract method is never consumed polymorphically (config.py:168-193). CommandStepConfig returns None; SimulationStepConfig uses _SUPPORTED_ARGS directly in _create_from_dict. No external code calls this method. It forces new step types to implement something unused. (Design — low severity)

ResourceConfig and NativeSpecification represent the same concept separately — manual field mapping in get_tasks() (config.py:553-563). Consider adding a ResourceConfig.to_native_specification(job_name) method to eliminate the manual mapping and keep the conversion in one place. (Design)

Maintainability
Hardcoded command="run" string instead of COMMANDS.run (config.py:569). The COMMANDS module is already imported; the rest of the codebase uses COMMANDS.run consistently. Suggestion: Change to command=COMMANDS.run. (Maintainability)

validate_runtime_and_queue("01:00:00", self.queue) magic sentinel (config.py:810). A throwaway runtime is passed to validate just the queue. A validate_queue() function would be clearer. (Maintainability)

_parse_steps() docstring says "Falls back to CommandStepConfig if type is not registered" but code raises ValueError (config.py:700-710). The docstring is stale. (Maintainability + Documentation)

ResourceConfig.to_dict() has dead is not None guards on non-optional memory_gb and runtime fields (config.py:109-121). These suggest nullability that doesn't exist. (Maintainability — low severity)

BaseStepConfig is not a @dataclass but defines __post_init__ (config.py:125-137). This works because subclasses are dataclasses, but a new subclass author forgetting @dataclass would get silent validation bypass. Add a comment or consider making the ABC a dataclass. (Maintainability)

collect_metadata() reconstructs job_specific dict by hand (writing.py:131-140) instead of using JobParameters.job_specific. If the job-specific fields ever change, this will silently drift. (Maintainability)

Inconsistent error types: WorkflowConfig raises KeyError for validation failures (duplicate step names, missing project) while all other validation uses ValueError. (Maintainability)

DRY
Duplicated "OutputPaths → keyspace → job params → task list" pipeline between SimulationStepConfig.get_tasks() (config.py:517-575) and runner.main() (runner.py:320-453). The core sequence of parsing branches, building job parameters, and creating tasks is repeated. The backup_configuration dict literal {"backup_dir": ..., "backup_freq": ..., "backup_metadata_path": ...} is assembled identically in both places. Suggestion: Extract the shared pipeline into a utility, or at minimum extract the backup config dict into a helper/TypedDict. (High risk — changes to the "build job parameters" contract must be applied in two places)

WorkflowConfig(...) boilerplate repeated 6 times in test_builder.py:87-171 across TestResourceDefaults and TestEnvironmentResolution. Suggestion: Add a make_single_step_workflow_config() helper to utilities.py. (Medium risk)

backup_configuration dict literal assembled as a raw dict in three places — config.py, jobs.py, and consumed by vivarium_work_horse.py. A stringly-typed contract with no shared type definition. Suggestion: Define a BackupConfig TypedDict or dataclass. (Medium risk)

Tests
SimulationStepConfig.get_tasks() is completely untested — the most complex method in the PR, containing OutputPaths creation, branch parsing, job parameter building, NativeSpecification construction, and get_task_list() delegation. None of this orchestration is exercised. (Critical gap)

WorkflowBuilder DAG wiring only tested with single-task steps (test_builder.py:69-96). The N×M cross-product wiring for fan-out steps is untested. (Significant gap)

Missing queue/project ValueError in get_tasks() untested (config.py:557-562).

Unsupported step type in _parse_steps() untested — no test for type: "bogus" raising ValueError.

_create_from_dict with missing required args (empty args dict) untested — Path(args["model_specification"]) would KeyError.

to_dict() round-trip for non-default backup_freq and sim_verbosity untested (config.py:590).

Load test result wrapping not verified — test_load_test_calls_load_test_work_horse (test_task_runner.py:127) doesn't assert write_task_results receives {"load_test": results_df}.

_build_argv test helper accepts unused worker_log_dir parameter (test_task_runner.py:38-48) — vestigial parameter that misleads readers.

Documentation
SimulationStepConfig not exported from __init__.py (__init__.py:14). Users importing from the package can't access the new class.

Concatenated string has missing space (config.py:556-557): "...are not configured properly." "Please check..." yields "properly.Please". Add a leading space.

CHANGELOG entry too vague (CHANGELOG.rst:1-3). The 3.2.0 entry doesn't mention type: simulation steps, SimulationStepConfig, or parallel task fan-out.

No user-facing documentation for psimulate workflow or the workflow YAML format. A new section in the docs covering the YAML schema, step types, and examples would be valuable.

SimulationStepConfig docstring omits backup_freq and sim_verbosity from the example args (config.py:459-477).

_create_from_dict() on SimulationStepConfig has only a one-line docstring while the parallel method on CommandStepConfig has full Parameters/Returns.

simulation_step_plan.md design document is stale — shows StepConfig as base class with different fields than the actual BaseStepConfig / CommandStepConfig / SimulationStepConfig hierarchy.

SUPPORTED_STEP_TYPES missing ClassVar annotation (config.py:638-640) — inconsistent with _SUPPORTED_ARGS which does use ClassVar.

Functionality
SimulationStepConfig.get_tasks() always passes command="run" to get_task_list() (config.py:569). This means simulation steps in workflows always behave as fresh runs. If a workflow step should ever support restart/expand semantics, this would need to change. Currently acceptable for the initial implementation, but worth noting.

write_task_results() inserts job-specific columns at position df.shape[1] - 1 (writing.py:79). For an empty DataFrame (0 columns), df.shape[1] - 1 = -1, which would raise IndexError. If a work horse returns an empty DataFrame for a metric, this would crash. Edge case worth guarding.

task_runner.main() adds loguru sinks unconditionally (task_runner.py:69-70) — logger.add(sys.stdout, level="INFO") and logger.add(sys.stderr, level="WARNING") are added without checking for existing sinks. In test scenarios or if main() is called multiple times, log lines would be duplicated. Minor in production (tasks are single-shot), but can cause noisy test output.

ResourceConfig.from_dict() uses data.get("project") or workflow_project (config.py:94). If the YAML explicitly sets project: "" (empty string), the or would fall through to workflow_project due to truthiness evaluation. This is likely the desired behavior but is an implicit conversion worth being aware of.

Minor Nits
make_simulation_step_dict() in utilities.py:73-87 puts model_specification and branch_configuration at the top level, not inside args — doesn't match the actual schema and would fail if used with _create_from_dict.

SUPPORTED_STEP_TYPES comment says "add new step types here" but has no attribute docstring for API docs.

Overall
The architecture is well-structured and the class hierarchy is a reasonable design for supporting multiple step types. The key areas for improvement are:

Test coverage — SimulationStepConfig.get_tasks() is the most complex and most important method in the PR, yet completely untested. This is the highest-priority gap.
OutputPaths creation inside get_tasks() creates a coupling between config objects and filesystem side effects, and the timestamp-based path generation may cause issues with resume or repeated builds.
CommandStepConfig silently drops hardware — a functional bug that could affect users.
DRY — the job-parameter-building pipeline and backup config dict are duplicated between the workflow config path and the runner path, creating maintenance risk.


tests/psimulate/workflow_config/test_config.py
```diff
@@ -235,3 +279,301 @@ def test_from_dict_all_fields(self) -> None:
...
+    def test_requires_required_fields(
+        self, missing_field: str, kwargs: dict[str, Any]
+    ) -> None:
+        """CommandStepConfig requires name and resources."""
+        with pytest.raises(TypeError, match=missing_field):
+            CommandStepConfig(**kwargs)
+
+
+class TestSimulationStepConfig:
+    """Tests for SimulationStepConfig - the simulation step type."""
```
albrja, line 332: All these tests are doing is verifying that if a user provides a workflow configuration:
`workflow:
    steps:
        name: p_sim
        type: simulation
        args:
             -config: /path/to/config.yaml
             -model_specification: /path/to/model_spec.yaml`

That a user has provided all the adequate arguments needed to for a proper "psimulate run" command we will infer from the configuration.

simulation_step_plan.md
```diff
@@ -0,0 +1,330 @@
+# Simulation Step Plan: Bespoke `type: "simulation"` in `psimulate workflow`
```
albrja: Can be ignored

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -30,10 +50,16 @@ class ResourceConfig:
 
     memory_gb: int
     """Memory in GB."""
+    project: str | None = None
```
albrja, line 53: The key feature update for the pr is in this file where we refactor the step config class so we can subclass with additional types moving forward and this pr implementing the simulation step. 

Steps now return a list of Jobmon tasks that will be submitted by the workflow.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,579 @@ def __post_init__(self) -> None:
...
+    """Configuration for a parallel simulation workflow step.
+
+    This step type directly creates parallel simulation tasks — one per
+    (input_draw, random_seed, branch) combination — rather than launching
+    ``psimulate run`` as a subprocess. The tasks use the same task runner
+    infrastructure as ``psimulate run``.
+
+    The ``resources`` field specifies compute resources for each individual
+    simulation task (memory, runtime, cores).
+
```
albrja, line 419: Note that we are not supporting the new feature to run the simulation step with a separate config file (psimulate run --config config.yaml. All necessary args and configuration will be in the workflow yaml configuration so (at least fo now) we do not have to parse nested yaml files and validate that the configuration for specific values exists in one of these places and which takes priority.

albrja:
I would highly encourage reviewers to look at changes before the last two commits and then after since a lot of the code-reviewer's feedback was refactoring code and DRY updates to methods used both by psimulate ru/restart and workflow.

Here is the summary of changes I made based on the code-reviewer's feedback:
PR Updates Log — PR #324: SimulationStepConfig
Tracking changes made in response to pr-review.txt feedback.

--- UPDATE 1: Stable build timestamp for workflow steps ----
Review item: Design #1

Problem:
  OutputPaths generated a new timestamp on every build() call, so resume
  builds produced different paths, splitting results across directories.

Solution:
  Builder generates a timestamp once, persists it to .build_timestamp, and
  passes it to every step via get_tasks(build_timestamp=...). On resume the
  persisted value is reused. This keeps all steps in one run sharing the
  same timestamp and makes output paths deterministic across restarts.

Files changed:
  - src/.../psimulate/paths.py — added optional launch_time param
  - src/.../workflow_config/config.py — added build_timestamp to get_tasks()
  - src/.../workflow_config/builder.py — added _get_or_create_build_timestamp()
  - tests/.../workflow_config/test_builder.py — mock fixture for timestamp
  - tests/.../workflow_config/test_config.py — updated get_tasks() call

--- UPDATE 2: CommandStepConfig respects hardware constraint ----
Review item: Design #2

Problem:
  CommandStepConfig.get_tasks() built compute_resources without the
  hardware constraint, silently dropping it even when the user specified it.

Solution:
  Added conditional that appends "constraints" to compute_resources when
  self.resources.hardware is set, matching the pattern in cluster/interface.py.

Files changed:
  - src/.../workflow_config/config.py — added hardware→constraints in get_tasks()

--- UPDATE 3: Early validation of queue/project ----
Review item: Design #3

Problem:
  queue/project were validated late inside SimulationStepConfig.get_tasks(),
  after config parsing and workflow building. Also had a missing space in the
  error string concatenation ("properly."Please → "properly. Please").

Solution:
  Moved the isinstance check to BaseStepConfig.__post_init__() so all step
  types fail fast at construction if queue/project aren't strings. Removed
  the redundant late check from get_tasks(). Improved the error message to
  suggest setting values at step or workflow level.

Files changed:
  - src/.../workflow_config/config.py — validation in __post_init__, removed from get_tasks()
  - tests/.../workflow_config/test_builder.py — added queue/project to ResourceConfig in fixtures
  - tests/.../workflow_config/test_config.py — added queue/project to ResourceConfig in step tests

--- UPDATE 4: Flatten from_dict/_create_from_dict indirection ----
Review item: Design #5

Problem:
  Two-layer from_dict() → _create_from_dict() existed solely to centralise
  a 3-line _validate_command_type_exclusivity check. Forced subclasses to
  implement an abstract _create_from_dict they'd never call directly.

Solution:
  Made from_dict() itself the abstract method on BaseStepConfig. Each
  subclass now implements from_dict() directly. Moved the command/type
  exclusivity check to _parse_steps() where routing already happens.
  Removed _validate_command_type_exclusivity and _create_from_dict.

Files changed:
  - src/.../workflow_config/config.py — flattened class hierarchy, moved check to _parse_steps()

--- UPDATE 5: ResourceConfig.to_native_specification() bridge method ----
Review item: Design #7

Problem:
  Both step types manually mapped ResourceConfig fields to Jobmon resources,
  with subtle inconsistencies: CommandStepConfig passed runtime as a string
  while SimulationStepConfig converted to seconds via NativeSpecification.
  CommandStepConfig also lacked stdout/stderr log routing.

Solution:
  Added to_native_specification(job_name) on ResourceConfig as the single
  conversion point. Both step types now go through NativeSpecification →
  to_jobmon_spec(), giving consistent runtime conversion, hardware
  constraints, and log routing. CommandStepConfig uses the workflow
  output_directory as its worker_logging_root. Further refined to build
  self.native_specification once in BaseStepConfig.__post_init__() so all
  step types get it automatically at construction time — get_tasks() just
  references self.native_specification.

Files changed:
  - src/.../workflow_config/config.py — added to_native_specification(), built in __post_init__()
  - tests/.../workflow_config/test_config.py — updated compute_resources assertions
  - tests/.../workflow_config/test_builder.py — updated resource assertions (float memory, seconds)

--- UPDATE 6: DRY up WorkflowConfig boilerplate in test_builder.py ----
Review item: DRY #2

Problem:
  Six tests across TestResourceDefaults and TestEnvironmentResolution each
  constructed a nearly-identical WorkflowConfig with one CommandStepConfig,
  varying only resources, default_environment, and step environment.

Solution:
  Extracted _make_single_step_config() helper that accepts keyword overrides
  for resources, default_environment, and step_environment, defaulting to the
  common values. All six tests now call the helper, eliminating ~60 lines of
  repeated construction boilerplate.

Files changed:
  - tests/.../workflow_config/test_builder.py — added helper, refactored 6 tests

--- UPDATE 7: BackupConfiguration TypedDict ----
Review item: DRY #3

Problem:
  The backup_configuration dict (backup_dir, backup_freq, backup_metadata_path)
  was assembled as a raw dict literal in jobs.py and config.py, and consumed
  via string keys in vivarium_work_horse.py — a stringly-typed contract with
  no shared definition.

Solution:
  Defined BackupConfiguration TypedDict in jobs.py (next to JobParameters which
  owns the field). Updated JobParameters.backup_configuration field type,
  build_job_list(), and SimulationStepConfig.get_tasks() to construct
  BackupConfiguration(...) instead of raw dicts. Consumer code in
  vivarium_work_horse.py unchanged (TypedDict supports [] access).

  Also fixed test_runner.py::test_write_configuration_workflow_command which
  was missing project/queue on ResourceConfig (pre-existing issue from the
  early-validation change in Update 3).

Files changed:
  - src/.../psimulate/jobs.py — added BackupConfiguration TypedDict, updated types
  - src/.../workflow_config/config.py — import + use BackupConfiguration
  - tests/.../test_runner.py — added project/queue to ResourceConfig

--- UPDATE 8: Test for SimulationStepConfig.get_tasks() ----
Review item: Tests #1

Problem:
  SimulationStepConfig.get_tasks() — the most complex method in the PR — had
  zero test coverage. OutputPaths creation, branch parsing, job parameter
  building, NativeSpecification wiring, and get_task_list() delegation were
  all untested.

Solution:
  Added test_get_tasks_wires_arguments which patches the four heavy
  dependencies (OutputPaths, Keyspace, build_job_parameters_from_keyspace,
  get_task_list) and verifies each is called with the correct arguments.
  Uses context-manager style patching per team convention.

Files changed:
  - tests/.../workflow_config/test_config.py — added test_get_tasks_wires_arguments

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -20,9 +21,92 @@
 from vivarium_cluster_tools.psimulate.results.writing import write_metadata
 
 if TYPE_CHECKING:
+    from jobmon.client.task import Task
     from jobmon.client.workflow import Workflow
 
 
+def get_task_list(
```
albrja, line 28: This is the start of the future API work we will do but it helped with the task now.

src/vivarium_cluster_tools/psimulate/results/writing.py
```diff
@@ -139,13 +139,7 @@ def collect_metadata(metadata_dir: Path, results_dir: Path) -> pd.DataFrame:
...
         with open(metadata_path) as f:
-            job_params = json.load(f)
-        # Build flattened job_specific dict matching what already_complete() expects
-        job_specific = {
-            **job_params.get("branch_configuration", {}),
-            "input_draw": job_params["input_draw"],
-            "random_seed": job_params["random_seed"],
-        }
-        row = dict(utilities.collapse_nested_dict(job_specific))
+            job_params = JobParameters(**json.load(f))
```
albrja, line 142: Small suggested change from code reviewer

src/vivarium_cluster_tools/psimulate/jobs.py
```diff
@@ -19,6 +19,14 @@
 from vivarium_cluster_tools.psimulate import branches
 
 
+class BackupConfiguration(TypedDict):
```
albrja, line 22: DRY

src/vivarium_cluster_tools/psimulate/cluster/interface.py
```diff
@@ -7,11 +7,130 @@
...
+    "r630",  # high capacity
+    "c6420v1",  # batch 1
+    "c6420v2",  # batch 2
+    "r650",  # high capacity
+    "r650v2",  # high capacity
+    "r650xs",  # high speed
+]
+
+
+def validate_project(project: str) -> str:
```
rmudambi: Why are all these methods in the interface module? They seem better situated in the cli_options module where they were before, or in a validation module.

src/vivarium_cluster_tools/psimulate/cluster/cli_options.py
```diff
@@ -12,47 +12,38 @@
...
+    QUEUE_MAX_RUNTIME_HOURS,
+    RUNTIME_FORMAT,
+    VALID_PROJECTS,
+)
+from vivarium_cluster_tools.psimulate.cluster.interface import (
+    validate_hardware as _validate_hardware_core,
+)
+from vivarium_cluster_tools.psimulate.cluster.interface import (
+    validate_runtime_and_queue as _validate_runtime_and_queue_core,
+)
```
rmudambi: These imports are weird. Why did we have to rename them?

albrja: Updated to remove the unnecessary wrapper and private name.

src/vivarium_cluster_tools/psimulate/cluster/interface.py
albrja: Made validation modlue

rmudambi (approved):

src/vivarium_cluster_tools/psimulate/cluster/validation.py
```diff
@@ -0,0 +1,142 @@
...
+
+def validate_project(project: str) -> str:
+    """Validate that *project* is a recognised cluster project.
+
+    Returns the project unchanged on success.
+
+    Raises
+    ------
+    ValueError
+        If *project* is not in ``VALID_PROJECTS``.
```
stevebachmeier, line 54: nit: I'm assuming you used claude for this PR? We typically use ` or `` for args, and this is using * throughout. We should get a handle on that for consistency's sake.

src/vivarium_cluster_tools/psimulate/cluster/validation.py
```diff
@@ -0,0 +1,142 @@
...
+def validate_runtime_and_queue(runtime: str, queue: str | None) -> tuple[str, str]:
+    """Validate and reconcile a runtime string with a queue.
+
+    * Validates runtime format (``hh:mm:ss``).
+    * If *queue* is ``None``, selects the shortest queue that can
+      accommodate *runtime*.
+    * If *queue* is given, checks that *runtime* does not exceed the
+      queue's maximum.
+
+    Returns the ``(runtime, queue)`` pair (queue may be resolved from
```
stevebachmeier: Put this into a proper Returns section. Possibly throughout all docstrings.

src/vivarium_cluster_tools/psimulate/cluster/validation.py
```diff
@@ -0,0 +1,142 @@
...
+    if not _RUNTIME_RE.match(runtime):
+        raise ValueError(
+            f"Invalid runtime '{runtime}'. " f"Expected format {RUNTIME_FORMAT}."
+        )
+
+    hours, minutes, seconds = runtime.split(":")
+    total_hours = int(hours) + float(minutes) / 60.0 + float(seconds) / 3600.0
+
+    max_runtime = max(QUEUE_MAX_RUNTIME_HOURS.values())
+    if total_hours > max_runtime:
```
stevebachmeier: This seems unnecessary since you do a queue-specific check below

src/vivarium_cluster_tools/psimulate/cluster/validation.py
```diff
@@ -0,0 +1,142 @@
...
+                f"Must be one of: {sorted(QUEUE_MAX_RUNTIME_HOURS)}."
+            )
+        if total_hours > QUEUE_MAX_RUNTIME_HOURS[queue]:
+            raise ValueError(
+                f"Runtime '{runtime}' exceeds the maximum for queue "
+                f"'{queue}' ({QUEUE_MAX_RUNTIME_HOURS[queue]}:00:00)."
+            )
+        return runtime, queue
+
+    # No queue specified — pick the shortest one that fits.
```
stevebachmeier: I think this is picking the first one that fits, not the shortest one

src/vivarium_cluster_tools/psimulate/cluster/cli_options.py
```diff
@@ -12,47 +12,34 @@
...
     "--project",
     "-P",
-    type=click.Choice(
-        [
-            "proj_simscience",
-            "proj_simscience_prod",
-            "proj_csu",
-        ]
-    ),
+    type=click.Choice(sorted(VALID_PROJECTS)),
```
stevebachmeier: nit: I wouldn't sort it b/c we basically never use proj_csu and so it's weird to have that first

src/vivarium_cluster_tools/psimulate/cluster/cli_options.py
```diff
@@ -98,64 +85,24 @@ def with_queue_and_max_runtime(func: CLIFunction) -> CLIFunction:
...
+    except ValueError as e:
+        raise click.BadParameter(str(e))
+    return value
 
 
 _with_queue = click.option(
     "--queue",
     "-q",
-    type=click.Choice(["all.q", "long.q"]),
+    type=click.Choice(sorted(QUEUE_MAX_RUNTIME_HOURS)),
```
stevebachmeier, line 105: QUEUE_MAX_RUNTIME_HOURS is a dict. I don't think this will work. Do you mean to add `.values()`?

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -20,9 +21,92 @@
...
+        "--command {command}"
+    )
+    if env is not None:
+        worker_command = f"conda run --no-capture-output -n {env} {worker_command}"
+
+    task_template = tool.get_task_template(
+        template_name="psimulate",
+        command_template=worker_command,
+        node_args=["task_id"],
+        task_args=["metadata_dir", "results_dir"],
```
stevebachmeier, line 85: Each task gets its own metadata_dir and results_dir? I thought these would be op_args but maybe not...

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -20,9 +21,92 @@
...
+
+    # Write job spec metadata (one JSON per task for the worker to pick up)
+    for job_params in job_parameters_list:
+        write_metadata(
+            metadata_dir=metadata_dir,
+            job_parameters=job_params,
+        )
+
+    # Batch-create all tasks
+    tasks = task_template.create_tasks(
```
stevebachmeier, line 99: This natively returns a list?

src/vivarium_cluster_tools/psimulate/results/writing.py
stevebachmeier: This seems different though. We were previously specifically adding `input_draw` and `random_seed` to the dict. I think that was a bugfix somewhere along the way, wasn't it @patricktnast?

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -10,50 +10,65 @@
...
-            default_cluster_name="slurm",
-        )
+    def build(self, workflow_args: str) -> Workflow:
+        """Build the full workflow DAG and return the Jobmon Workflow.
 
+        Parameters
+        ----------
+        workflow_args
+            Deterministic string that Jobmon uses to identify the workflow.
+            Must be identical across runs for resume to work.
```
stevebachmeier, line 54: Add Returns sections for all public methods

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -66,27 +81,51 @@ def build(self) -> Workflow:
...
+                build_timestamp=build_timestamp,
             )
-            tasks.append(task)
 
-        # Wire sequential dependencies: each step depends on the previous
-        for i in range(1, len(tasks)):
-            tasks[i].add_upstream(tasks[i - 1])
+            # Wire sequential dependencies: every task in this step
+            # depends on every task from the previous step.
+            for task in step_tasks:
```
stevebachmeier, line 93: Ok so we still assume each step is sequential and the new thing is that any of those steps could actually call psimulate?

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -66,27 +81,51 @@ def build(self) -> Workflow:
...
+
+        Returns
+        -------
+            Timestamp string in ``YYYY_MM_DD_HH_MM_SS`` format.
+        """
+        timestamp_path = self.config.output_directory / BUILD_TIMESTAMP_FILENAME
+        if timestamp_path.exists():
+            return timestamp_path.read_text().strip()
+
+        build_timestamp = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
```
stevebachmeier, line 128: I personally prefer not having a million underscores but maybe it's just me, i.e. I prefer `20260428095830` or `20260428_095830` to `2026_04_28_09_58_30`

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -10,18 +10,38 @@
...
+    from jobmon.client.task import Task
+
 REQUIRED_WORKFLOW_FIELDS = {"name", "steps"}
 
 DEFAULT_MAX_ATTEMPTS = 2
 
-VALID_PROJECTS = {"proj_simscience", "proj_simscience_prod"}
-VALID_QUEUES = {"all.q", "long.q"}
+
+DEFAULT_BACKUP_FREQ_SECONDS = 30.0 * 60.0
```
stevebachmeier, line 43: Can we extract the actual frequency instead of hard-coding?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,541 @@ def __post_init__(self) -> None:
...
+        *,
+        workflow_project: str | None = None,
+        workflow_queue: str | None = None,
+    ) -> ResourceConfig:
+        """Create a ResourceConfig from a dictionary.
+
+        Step-level values take precedence; workflow-level defaults fill in
+        any that are absent.
+
+        Parameters
```
stevebachmeier, line 93: add returns sections everywhere

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,541 @@ def __post_init__(self) -> None:
...
 
+    def to_native_specification(self, job_name: str) -> NativeSpecification:
+        """Convert to a :class:`~vivarium_cluster_tools.psimulate.cluster.interface.NativeSpecification` for Jobmon task submission.
+
+        Parameters
+        ----------
+        job_name
+            The SLURM job name for this step's tasks.
+        """
+        if not isinstance(self.project, str) or not isinstance(self.queue, str):
```
stevebachmeier, line 137: This feels like a weird check. I mean, it's true, but what makes this special? We don't always check types like this.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,541 @@ def __post_init__(self) -> None:
...
+            max_runtime=self.runtime,
+            hardware=self.hardware or [],
+        )
+
+
+class BaseStepConfig(ABC):
+    """Abstract base class for all workflow step configurations.
+
+    Defines the interface that all step types must implement. Concrete
+    subclasses **must** be decorated with ``@dataclass`` so that
```
stevebachmeier, line 156: Is this just to dangerous? Is there a way to confirm that concrete classes are actually dataclasses?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,541 @@ def __post_init__(self) -> None:
...
+    environment: str | None
+
+    def __post_init__(self) -> None:
+        """Common validation for all step types, then call subclass validation.
+
+        This method is called automatically by @dataclass after __init__.
+        It performs validation common to all steps, then dispatches to the
+        subclass-specific _validate() method.
+        """
+        if not hasattr(self, "__dataclass_fields__"):
```
stevebachmeier, line 176: This is useless here since post_init only applies to dataclasses. If you move this into init, then it would actually alleviate my concern in the comment above

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -42,73 +68,541 @@ def __post_init__(self) -> None:
...
+        -------
+            Dictionary representation of the step configuration.
+        """
+        pass
+
 
 @dataclass
-class StepConfig:
-    """Configuration for a single workflow step."""
+class CommandStepConfig(BaseStepConfig):
```
stevebachmeier, line 307: I'm glad to see you are making different classes. I never really loved the implementation from the last PR

src/vivarium_cluster_tools/psimulate/cli.py
```diff
@@ -444,6 +444,12 @@ def test(
     default=None,
     help="Override maximum Jobmon task attempts from config file.",
 )
+@click.option(
+    "--resume",
```
stevebachmeier, line 448: This feels random for this PR. That said...should it be `restart` since that's what `psimulate` uses?

stevebachmeier (changes requested):
I'm mostly concerned about the small change in writing.py

src/vivarium_cluster_tools/psimulate/cluster/validation.py
albrja: I did not use claude but I will update the docstrings

src/vivarium_cluster_tools/psimulate/cluster/validation.py
albrja: This was a fail fast but is still redundant so I have removed it.

src/vivarium_cluster_tools/psimulate/results/writing.py
stevebachmeier: I'll approve with the assumption you'll manually confirm that this is indeed the same as before

stevebachmeier (approved):

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This is to appease mypy

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: See comment below

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This is the best way I found to handle this. It is a result of using ABC and dataclass. We have to do this check after the subclass (dataclass instance) is created so it has to be in __post_init__ . It is kind of annoying but it was the best way I found to do this.

PR #325, opened 2026-04-29
Albrja/mic 6961/slack user

albrja:
### Add helper method and piping to slack users when workflows finish
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6961

### Changes and notes
--add method and pipe into runners to slack users when workflows finish

### Code reviewer
PR Review: MIC-6960 — Slack Notifications on Workflow Completion
Summary
This PR adds Slack DM notifications to psimulate. When any workflow completes (via psimulate run, restart, expand, or workflow), the launching user receives a Slack DM indicating success or failure. The implementation consists of a new notifications.py module, a refactor of _bind_and_run_workflow() to return a (status, monitoring_url) tuple, and call sites in both workflow_main() and main().

Design
No HTTP timeouts — can hang the process indefinitely — notifications.py:41-79: All three requests.post() calls have no timeout. If Slack is unreachable, the post-workflow process hangs forever, holding SLURM resources. Add timeout=10 to each call.

Notification is unreachable when workflow.run() raises — runner.py:138-144: If _bind_and_run_workflow raises (submission failure, wf_status is None), send_slack_notification is never called. The user gets no notification in the scenario where they most need one. Consider wrapping in try/finally at the call site.

chat.postMessage response is silently discarded — notifications.py:76-79: The first two API calls check resp.json().get("ok") and log on failure. The final call discards its response entirely. This is inconsistent — if the message fails to post, there's no diagnostic trace.

Maintainability
Hard-coded @uw.edu email domain — notifications.py:39: No configuration escape hatch. If a user's Slack email differs from {USER}@uw.edu, notifications silently fail with no way to override. Consider an env var like PSIMULATE_SLACK_EMAIL as an override.

monitoring_url uses empty string "" as sentinel instead of None — runner.py:82-83: _bind_and_run_workflow returns "" when no GUI URL, but send_slack_notification accepts str | None. Works via truthiness, but the semantic mismatch may confuse readers. Return None instead for consistency with the downstream type.

Magic string "D" used without constant — notifications.py:66: The Jobmon status code "D" means DONE but this isn't self-evident. Also appears in runner.py:145. Consider a named constant.

Tests
No test for missing token (early return) — This is the most common production path. If someone accidentally removes the guard, notifications would crash for all users without a token. Add:

def test_no_token_skips_notification(monkeypatch):    monkeypatch.delenv("PSIMULATE_SLACK_BOT_TOKEN", raising=False)    with patch("...requests.post") as mock_post:        send_slack_notification(workflow_name="w", status="D")        mock_post.assert_not_called()
No test for Slack API failure (graceful degradation) — users.lookupByEmail returning {"ok": false}, conversations.open failure, and network exceptions are all untested. These protect against regressions that would surface as uncaught exceptions.

No test for monitoring_url=None / results_dir=None — The conditional message lines are never exercised with None values. The message-building logic should be verified to omit those lines.

Brittle assertion style — assert "testuser@uw.edu" in str(lookup_call) stringifies the entire call object. Prefer asserting on the specific data kwarg: assert lookup_call[1]["data"] == {"email": "testuser@uw.edu"}.

Documentation
Missing Parameters section on send_slack_notification() — notifications.py:22-28: Callers can't tell what status="D" means without reading the implementation. Document the expected values.
Functionality
Binary status mapping loses information — notifications.py:66-67: "DONE" if status == "D" else "ERROR" maps every non-"D" status to "ERROR". Jobmon has aborted/halted states that aren't really "errors." Consider including the raw status in the message (e.g., "ERROR (status='A')") so users can correlate with Jobmon.
Overall
Clean, focused implementation that follows the TDD plan well. The graceful degradation pattern (try/except, check API responses) is good. Priority items:

Add timeout=10 to all requests.post() calls — highest risk; can hang the process
Add the missing "no token" test — protects the most common path
Check chat.postMessage response — consistency with the other two calls
Remove # mypy: ignore-errors — no apparent reason for it

### Testing
image


src/vivarium_cluster_tools/psimulate/notifications.py
```diff
@@ -0,0 +1,82 @@
...
+            headers=headers,
+            data={"email": email},
+        )
+        lookup_data = lookup_resp.json()
+        if not lookup_data.get("ok"):
+            logger.warning(
+                f"Slack user lookup failed for {email}: {lookup_data.get('error')}"
+            )
+            return
+        user_id = lookup_data["user"]["id"]
```
albrja, line 61: Note: This is currently just getting the users id from slurm and slacking them. We can support slacking to channels, all we would have to do is add the bot to a channel but I just wanted to get something up and running. We would have to add in minor logic to get user id vs channel name.

src/vivarium_cluster_tools/psimulate/notifications.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Open a DM conversation with the user
+        convo_resp = requests.post(
```
rmudambi, line 64: Do we want this to go to a DM, or should we have a dedicated slack channel for this (like our CI channels), and ping the appropriate user?

src/vivarium_cluster_tools/psimulate/paths.py
```diff
@@ -218,7 +223,10 @@ def from_entry_point_args(
         elif command == COMMANDS.load_test:
             output_directory = output_directory / "load_test" / launch_time
 
-        logging_directory = output_directory / "logs" / f"{launch_time}_{command}"
+        logging_timestamp = (
```
albrja, line 226: This and the "is_resume" argument being passed around is a small bug I ran into testing restart/resume of workflow and keeping the logging directory structure consistent with the existing psimulate restart

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -101,6 +103,15 @@ def build(self, workflow_args: str) -> Workflow:
 
         return workflow
 
+    def _is_resume(self) -> bool:
```
stevebachmeier, line 106: how would this work w/ expand?

src/vivarium_cluster_tools/psimulate/notifications.py
stevebachmeier: I like the idea of a DM to keep noise reasonable. If we send to a (configurable?) channel, then we should defiitely tag the person

stevebachmeier (changes requested):
I'd just like clarification that this works w/ restart as well

Edit: I meant works with expand, not restart

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
albrja: Not currently

stevebachmeier:
I also note that I can squash and merge this thing despite it not having any reviews. Make sure you turn the merge ruleset back on!

edit: It was on but didn't have all of the regex branch names we normally do. I've added them.

albrja:
@stevebachmeier this does work with restart. Is there something specific you would like to see?

stevebachmeier:
> @stevebachmeier this does work with restart. Is there something specific you would like to see?

Oh I meant expand

albrja:
We are not supporting expand with workflows at the moment

stevebachmeier (approved):
Hard to really read, but as long as it works then 👍 . It is kinda weird that it will work for run and restart but not expand though. Are we sure we want to implement this with that pretty big gap?

rmudambi (approved):

PR #326, opened 2026-05-04
Make VCT require slurm node resources weekly

patricktnast:
### Description

- *Category*: 
bugfix
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7051

### Changes and notes
 
add the "weekly" flag to VCT so that we build end to end tests appropriately.
also, replace raise for unset env var with a default value.
### Testing

ran with jenkins, with run weekly flag and without

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -107,15 +107,15 @@ def get_workflow_timeout_seconds() -> int:
     Raises
     ------
     RuntimeError
-        If ``SLURM_JOB_ID`` is not set, the remaining time is less than the safety
+        If the remaining time is less than the safety
         buffer, or the remaining time cannot be determined from ``squeue``.
     """
     job_id = os.environ.get("SLURM_JOB_ID")
     if job_id is None:
-        raise RuntimeError(
-            "SLURM_JOB_ID is not set. psimulate must be run from within a "
-            "SLURM allocation (e.g. via srun)."
-        )
+        logger.info("SLURM_JOB_ID is unset. The workflow is likely being run"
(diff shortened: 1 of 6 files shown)


src/vivarium_cluster_tools/psimulate/cluster/interface.py
```diff
@@ -89,6 +89,10 @@ def _runtime_to_seconds(runtime_str: str) -> int:
         return _parse_slurm_time(runtime_str)
 
 
+# Default timeout in seconds for the jobmon workflow when no SLURM allocation
+# is detected (e.g. running from Jenkins).  10 hours.
+JOBMON_DEFAULT_TIMEOUT = 36000
```
rmudambi: I assume we can't import this from jobmon directly?

rmudambi (approved):

albrja (approved):

tests/psimulate/test_e2e.py
```diff
@@ -44,7 +44,7 @@
 
 RESULTS_DIR = "/mnt/team/simulation_science/priv/engineering/tests/output/"
 
-pytestmark = [pytest.mark.cluster, pytest.mark.slow, pytest.mark.weekly]
+pytestmark = [pytest.mark.cluster, pytest.mark.slow]
```
albrja: I assume if I have a e2e file with the same marker will I just be able to remove the weekly marker as well and it will run via CI once a week?

patricktnast: oh, actually, i need to revert this

src/vivarium_cluster_tools/psimulate/cluster/interface.py
patricktnast: youd have to inspect it from the workflow.run method, but you could also just not pass in anything, actually

PR #327, opened 2026-05-05
Albrja/mic 7001/pytest step

albrja:
### Add custom pytest step type
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7061

### Changes and notes
-add pytest custom step type

### Testing
Successfully ran workflow with new custom step type

### Code reviewer
PR Review: Add PytestStepConfig step type (MIC-7001)
Summary
This PR introduces a PytestStepConfig class — a new structured step type for the psimulate workflow pipeline that constructs and executes a pytest command as a single Jobmon task. Users can define pytest steps in their workflow YAML with structured args (path, k, runslow, numprocesses) instead of raw command strings, gaining cross-field validation (e.g., numprocesses <= cores). The class is registered in WorkflowConfig.SUPPORTED_STEP_TYPES and exported from the package __init__.py.

Design
Shell injection risk in _build_command() — config.py:698-709: User-supplied self.path is interpolated directly into a shell command without quoting. While self.k gets double-quoted, self.path does not. A path with spaces or shell metacharacters (e.g., from a template source) would break or potentially allow unintended execution. Use shlex.quote() for both self.path and self.k:

import shlexdef _build_command(self) -> str:    parts = ["pytest"]    if self.path:        parts.append(shlex.quote(self.path))    if self.k:        parts.append(f"-k {shlex.quote(self.k)}")    if self.runslow:        parts.append("--runslow")    if self.numprocesses > 1:        parts.append(f"-n {self.numprocesses}")    return " ".join(parts)
runslow hard-codes a project-specific flag — config.py:647: --runslow is a custom conftest.py convention, not a standard pytest flag. The next custom flag (e.g., --runintegration) requires another PR touching 5 sync points. Consider adding an extra_args: list[str] escape hatch for arbitrary flags while keeping runslow as a convenience shortcut. This would future-proof the design without requiring a full refactor each time.

Resource waste not surfaced — config.py:651: When cores: 4 but numprocesses is omitted (defaults to 1), the system allocates 4 cores but uses 1 for serial pytest. This is defensible (explicit > implicit) but could confuse users. Consider a logged warning when cores > 1 and numprocesses == 1.

Maintainability
Test/implementation mismatch is a bug — config.py:709 uses f"--numprocesses {self.numprocesses}" but test_config.py:738 asserts '-n 4'. One of them is wrong. The short form -n is the pytest-xdist convention most developers recognize. Recommend updating the implementation to -n (or fixing the test to expect --numprocesses).

to_dict() / dataclass default coupling is implicit — config.py:711-728: to_dict() omits numprocesses when == 1 and runslow when False. If someone changes a dataclass default but forgets to update to_dict(), round-tripping silently breaks. Add a comment noting this coupling or define defaults as named constants referenced in both places.

make_pytest_step_dict does shallow dict.update() — utilities.py:153: Calling make_pytest_step_dict(args={"k": "foo"}) replaces the default args dict entirely (losing "path": "tests/"). A test author expecting additive behavior will get confusing failures. Either document this clearly or implement merge behavior for nested args.

DRY
get_tasks() duplication (High severity) — config.py:670-696 is identical to config.py:345-366 except for self._build_command() vs self.command. Extract a _create_single_command_task(tool, env, command) helper in BaseStepConfig:

def _create_single_command_task(self, tool: Tool, *, env: str, command: str) -> Task:    task_template = tool.get_task_template(        template_name="workflow_command_step",        command_template="conda run --no-capture-output -n {env} {command}",        node_args=["command"], task_args=[], op_args=["env"],        default_cluster_name="slurm",    )    compute_resources = self.native_specification.to_jobmon_spec(        worker_logging_root=self.output_directory,    )    return task_template.create_task(        name=self.name, compute_resources=compute_resources, env=env, command=command,    )
from_dict() arg-validation boilerplate (Medium severity) — The 8-line block (get args, compute unsupported set, raise ValueError) is copy-pasted between SimulationStepConfig and PytestStepConfig. Extract as BaseStepConfig._validate_args(cls, data) classmethod.

to_dict() envelope (Low severity) — The name/type/resources/environment boilerplate is repeated but only has 2 consumers with slightly different shapes. Acceptable for now; revisit if more step types are added.

Tests
Bug: test_get_tasks_builds_correct_command will fail — test_config.py:738 expects -n 4 but implementation produces --numprocesses 4. Either the implementation or the test assertion must be fixed.

Missing: No test for minimal command (path only) — The most common usage (path="tests/", no k, no runslow, numprocesses=1) has no test verifying the output is "pytest tests/". Add:

def test_build_command_minimal_path_only(self) -> None:    config = PytestStepConfig(...)    assert config._build_command() == "pytest tests/"
Missing: No test for k-only command (no path) — No test verifies output when path=None, k="some_expr". A regression adding None to the command would be undetected.

Missing: No test for numprocesses=1 omission from command — The > 1 condition in _build_command() is never explicitly tested to confirm the flag is absent when numprocesses is 1.

Missing: No round-trip test — to_dict() → from_dict() round-trip would catch asymmetric serialization bugs as fields are added.

Weak assertion — test_config.py:694-700: test_to_dict_omits_unset_optional_fields only asserts absence of optional fields but doesn't verify path IS present. Strengthen with assert result["args"] == {"path": "tests/"}.

Documentation
path and k field docstrings mislead about optionality — config.py:644-645: Both say "Optional" but _validate enforces at least one. Change to: """Test path (file or directory). At least one of pathork is required."""

from_dict lacks Parameters/Returns section — config.py:735: CommandStepConfig.from_dict has full numpydoc sections. PytestStepConfig.from_dict only has a one-line summary. Match the more complete convention.

CHANGELOG doesn't mention the new step type — CHANGELOG.rst:3: Add a bullet: "Add ``pytest`` step type for structured test execution in workflows".

Class docstring should mention the path/k requirement — config.py:603-629: Add a note before the example: "At least one of path or k must be provided in the args section."

Functionality
_build_command() / test mismatch is a correctness bug — The implementation emits --numprocesses but the test expects -n. Since this is the only test exercising the full command output, either (a) the test was written against a different version of _build_command() or (b) the implementation was changed after the test was written. The test is the source of truth for intended behavior — update the implementation to use -n.

_validate() uses falsy check for self.path — config.py:653: if not self.path and not self.k means path="" passes validation but then _build_command() skips it (since if self.path: is also falsy for empty string). An empty-string path is semantically invalid but silently accepted if k is also provided. Consider explicit None checks instead: if self.path is None and self.k is None.

-k quoting strategy is fragile — config.py:703: The current approach wraps in double quotes (f'-k "{self.k}"'). If a user's -k expression itself contains double quotes (e.g., k: 'test_func[param="val"]'), the quoting breaks. Using shlex.quote() handles this correctly.

Minor Nits
config.py:709: The numprocesses > 1 condition in both _build_command() and to_dict() is repeated — could be a tiny helper @property def _is_parallel(self) -> bool but not worth it with only 2 uses.

config.py:651: The numprocesses field docstring mentions both -n and --numprocesses but the implementation only generates one form. Simplify the docstring to match whichever form is chosen.

Overall
The PR is well-structured and follows the established patterns in the codebase. The PytestStepConfig abstraction is justified — it provides meaningful cross-field validation that a raw CommandStepConfig cannot.

Key items to address before merge (in priority order):

Fix the --numprocesses / -n mismatch between implementation and test (this is a correctness bug)
Add shlex.quote() to _build_command() for shell safety
Extract _create_single_command_task() helper to eliminate the get_tasks() code duplication
Add missing test cases for minimal commands (path-only, k-only) and numprocesses=1 omission


src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -294,6 +296,40 @@ def get_tasks(
         """
         pass
 
+    def _create_single_command_task(self, tool: Tool, *, env: str, command: str) -> Task:
```
albrja, line 308: DRY

tests/psimulate/workflow_config/test_config.py
```diff
@@ -582,3 +584,223 @@ def test_get_tasks_wires_arguments(
 
             # -- Assert: returns whatever get_task_list returns --
             assert result is sentinel_tasks
+
+
+class TestPytestStepConfig:
+    """Tests for PytestStepConfig - the pytest step type."""
+
+    def test_supported_arguments_returns_expected_set(self) -> None:
```
albrja, line 592: I don't know if people thoroughly look through the tests but I constantly feel like there is a lot of test code because for each class I have to verify that:
-supported args are in fact supported 
-no extra args can be passed in
-special cases for class specific args that usually have weird edge cases
-configuration handling and return since we write out configuration now

This isn't a lot to check but it does cause there to be a lot of code and I am not sure how to condense it so there are different arguments and use cases for each subclass.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -601,13 +619,188 @@ def from_dict(
...
+        steps:
+          - name: unit_tests
+            type: pytest
+            resources:
+              memory_gb: 8
+              runtime: "01:00:00"
+              cores: 4
+            args:
+              path: tests/
+              k: "not slow"
```
rmudambi: What does "not slow" mean? It is confusingly similar to --runslow

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -601,13 +619,188 @@ def from_dict(
...
+    """Optional environment name to use for this step."""
+    path: str | list[str] | None = None
+    """Test path(s) (file or directory) to pass to pytest. Can be a single string
+    or a list of strings. At least one of ``path`` or ``k`` is required."""
+    k: str | None = None
+    """Pytest ``-k`` expression to filter tests by name. At least one of ``path`` or ``k`` is required."""
+    runslow: bool = False
+    """Whether to pass --runslow flag."""
+    numprocesses: int = 1
+    """Number of parallel workers for pytest-xdist (``-n``/``--numprocesses``). Defaults to 1 (no parallelism). Must be <= resources.cores."""
```
rmudambi: I thought we were going to exclude this argument and just always use `resources.cores`. Why would a user ever not want to use all the cores they requested?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -601,13 +619,188 @@ def from_dict(
...
+        if self.resources.cores > 1 and self.numprocesses < self.resources.cores:
+            logger.warning(
+                f"Step '{self.name}': resources.cores is {self.resources.cores} but "
+                f"numprocesses is {self.numprocesses}. The extra cores will be allocated "
+                f"but unused. Set numprocesses to utilize them or reduce cores."
+            )
+
+    def supported_arguments(self) -> set[str]:
+        """Return valid keys for the 'args' section of pytest steps."""
+        return self._SUPPORTED_ARGS
```
rmudambi: This is the same as in the `SimulationStepConfig`. Why can't we move this logic up into the base class, rather than have it be an abstract method there?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -601,13 +619,188 @@ def from_dict(
...
+        tool: Tool,
+        *,
+        env: str,
+        build_timestamp: str,
+        is_resume: bool = False,
+    ) -> list[Task]:
+        """Create a single Jobmon Task for this pytest step."""
+        return [
+            self._create_single_command_task(tool, env=env, command=self._build_command())
+        ]
```
rmudambi: This is identical to the CommandStepConfig method. You should define it in the base class, and have SimulationStepConfig override it. You may not need to extract the helper method if you do this.

tests/psimulate/workflow_config/test_config.py
rmudambi: You could try asking Claude if there are ways to leverage helper functions or test parameterization to reduce duplication.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This was just a terrible filter example.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: That is fair.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: Yea this will also help with the python step next.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -166,6 +169,10 @@ class BaseStepConfig(ABC):
     output_directory: Path
     environment: str | None
 
+    _SUPPORTED_ARGS: ClassVar[set[str] | None] = None
+    """Subclasses override this with a set of valid 'args' keys, or leave as
```
stevebachmeier: This docstring doesn't actually say what it is.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -204,7 +211,6 @@ def _validate(self) -> None:
         """
         pass
 
-    @abstractmethod
     def supported_arguments(self) -> set[str] | None:
```
stevebachmeier, line 215: seems like this should be a property

stevebachmeier: of if you keep it a method, rename to `get_supported_arguments`

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -255,7 +258,6 @@ def from_dict(
         """
         pass
 
-    @abstractmethod
     def get_tasks(
```
stevebachmeier, line 262: this kind of feels like it should be a `tasks` property as well

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
stevebachmeier: Why all of a sudden are there these required kward separators in the signatures out of curiousity? I don't hate them, but like...why and how are you choosing where to insert it? e.g. why isn't `tool` a required kwarg but `env` and `command` are? All three are very obvious what they are by their assignment so that typical heuristic doesn't seem to apply

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -255,7 +258,6 @@ def from_dict(
         """
         pass
 
-    @abstractmethod
```
stevebachmeier, line 258: why isn't this abstract anymore?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -478,9 +493,11 @@ def _validate(self) -> None:
                 f"Step '{self.name}': simulation type requires 'branch_configuration'."
             )
 
-    def supported_arguments(self) -> set[str]:
-        """Return valid keys for the 'args' section of simulation steps."""
-        return self._SUPPORTED_ARGS
+    def _build_command(self) -> str:
```
stevebachmeier, line 497: If this isn't used then why make it abstract?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -626,6 +625,12 @@ class PytestStepConfig(BaseStepConfig):
     This step type constructs a ``pytest`` command from structured arguments
     and runs it as a single Jobmon task.
 
+    When ``resources.cores`` is greater than 1, the generated command includes
+    ``--numprocesses <cores>`` to enable parallel test execution via
+    `pytest-xdist <https://pypi.org/project/pytest-xdist/>`_.  The test
+    environment **must** have ``pytest-xdist`` installed when ``cores > 1``;
+    otherwise pytest will fail with an "unrecognized arguments" error.
```
rmudambi: You don't need to tell the user how we're doing this under the hood. All they need to know is that we will parallelize their tests across the requested threads. We also don't need to tell them they need to install pytest-xdist - that should be a dependency of VCT at this point.

rmudambi (approved):

stevebachmeier (approved):

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This takes parameters.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: Because all the "command" steps can just use the same implementation.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: This is the only class that doesn't use this.

PR #329, opened 2026-05-11
Albrja/mic 7063/v2 notebook step

albrja:
### Implementation of notebook step type
- *Category*: Implementation
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7063

### Changes and notes
-implementation for notebook step type

### Code reviewer
PR Review: Add NotebookStepConfig to Workflow System
Summary
This PR introduces NotebookStepConfig, a new step type for the psimulate workflow system that allows users to execute Jupyter notebooks via papermill as workflow steps. The implementation follows the existing BaseStepConfig pattern with a @dataclass subclass that validates configuration, builds a papermill shell command, and supports serialization/deserialization via to_dict()/from_dict(). It includes comprehensive unit tests covering validation, command building, round-trip serialization, and YAML integration. Shared validation logic (_validate_scalar_dict) was extracted as a module-level utility reusable by both NotebookStepConfig and PythonStepConfig.

Design
Bug: Incorrect papermill -y flag usage in _build_command() — config.py:1114-1119

The code emits -y {key} {yaml_value} (e.g., -y verbose true), but papermill's -y/--parameters_yaml flag takes a single YAML string argument, not a name-value pair. With click's multiple=True, -y verbose parses "verbose" as the YAML snippet and true becomes a stray positional argument, causing a runtime error.

Fix:

yaml_snippet = f"{key}: {yaml_value}"parts.append(f"-y {shlex.quote(yaml_snippet)}")
This produces -y 'verbose: true', which papermill correctly parses as {verbose: True}.

_IDENTIFIER_RE allows hyphens, which are invalid Python identifiers — config.py:56

The shared regex ^[a-zA-Z0-9_-]+$ permits keys like my-param. Papermill injects parameters as Python variable assignments, so hyphens cause SyntaxError at notebook execution time. The regex is intentionally permissive for PythonStepConfig (where --my-param is valid), so NotebookStepConfig._validate() should add its own stricter check (e.g., ^[a-zA-Z_][a-zA-Z0-9_]*$).

Promoted dataclass fields (vs. opaque args dict) is the right choice. This is consistent with SimulationStepConfig and PytestStepConfig. PythonStepConfig's opaque self.args dict is the outlier, not NotebookStepConfig. The promoted fields provide type safety, IDE support, and clear required-vs-optional semantics.

parameters type hint could be narrowed — config.py:1083. dict[str, Any] could be dict[str, str | int | float | bool | None] to make the validated invariant visible to type checkers. Minor improvement.

Maintainability
Hardcoded kernel name "python3" with no override — config.py:1100. The class docstring states it "exposes a notebook-agnostic schema so the executor can be swapped later," but -k python3 is a magic string. Extract to a class-level constant (e.g., _DEFAULT_KERNEL: ClassVar[str] = "python3") so it's discoverable and documented as intentionally not yet configurable.

Raw KeyError when output_path is missing from YAML args — config.py:1153. _build_from_dict accesses args["output_path"] unconditionally. A user omitting output_path gets a bare KeyError with no step name or guidance, unlike every other validation path. Either guard with .get() + explicit ValueError, or validate presence before accessing.

-k python3 is a single string element — config.py:1100. Elsewhere flags and values are separate entries or individually shlex.quote-d. Use "-k", shlex.quote("python3") for consistency. This prevents a latent injection risk if the kernel becomes configurable.

DRY
PythonStepConfig._validate_scalar partially duplicates _validate_scalar_dict — Both construct the same allowed tuple and produce the same error format. Extract a shared _check_scalar(value, label, step_name, allow_none) module-level function that both _validate_scalar_dict and _validate_positional_args call. This eliminates the duplicate logic while keeping the different call sites clean.

to_dict() envelope and _build_from_dict() resource-construction are repeated across all five step types but are small enough (~5 lines each) that extracting them now would add more indirection than value. Worth noting for when a sixth step type arrives.

Tests
No test for _build_from_dict when output_path is missing — The dataclass constructor tests catch missing fields via TypeError, but _build_from_dict would raise a raw KeyError before reaching dataclass validation. Add a test documenting this behavior (and ideally fix the implementation to produce a clear ValueError).

No float parameter type tested in _build_command — The parametrized test_build_command covers str, int, bool, None, but not float. Add a case like {"threshold": 0.05} to lock in formatting.

to_dict omission of optional fields is untested — test_to_dict_serialization sets all optional fields. There's no companion test verifying that parameters, cwd, and environment are omitted when unset.

Round-trip test doesn't assert cwd or environment — test_config.py:1264-1283. A bug in cwd/environment deserialization could go undetected.

No E2E test for notebook steps — test_workflow_e2e.py tests command, python, and command steps but no notebook step. This is the highest-risk coverage gap.

No adversarial-input test for shlex.quote safety — Parameter values go through shlex.quote, but no test verifies this with values like $(rm -rf /). This is a security-relevant code path worth pinning.

make_notebook_step_dict default is incomplete — utilities.py:137-154. Missing output_path in default args means the helper can't be used bare with from_dict(), unlike sibling helpers.

Documentation
Design document is out of date — pipeline_automation_plan.md:217 describes a flat notebook schema with auto-inferred output_path ({output_directory}/executed/<name>), but the implementation requires output_path explicitly in a nested args dict. This would mislead anyone referencing the design doc to write pipeline YAML.

Class docstring and field docstrings are accurate and well-written. The YAML example correctly reflects the nested args schema, and the Notes on parameter mapping are clear.

Functionality
_build_command() produces an invalid papermill command for bool/None parameters (see Design #1). This will cause runtime failures for any notebook step that uses bool or None parameter values. The str/int/float paths work correctly since they use -p which does take NAME VALUE pairs.

Hyphened parameter keys will silently pass validation but fail at notebook execution time (see Design #2). Users would get a confusing SyntaxError from the notebook kernel rather than an upfront validation error.

_build_from_dict does not validate that path key exists in args either — config.py:1152. Both path and output_path are accessed unconditionally. A YAML missing args.path would produce a bare KeyError with no context.

cwd defaulting to self.path.parent is correct — but when path is relative (before _build_from_dict resolves it), the parent might not be meaningful. Since _build_from_dict calls Path(args["path"]).resolve() before constructing the dataclass, this is fine in practice. The direct-construction case (without from_dict) could theoretically pass a relative path with no cwd override, making the cwd effectively relative too — but this seems like an acceptable edge case.

Minor Nits
config.py:1089 — The _validate method's f-string f"got {self.path!r}." uses an f-string concatenation (f"..." f"...") that could be a single f-string for readability.
Overall
The overall design is sound and consistent with the existing step type family. The most critical issue is the incorrect papermill -y flag usage (Design #1), which is a runtime correctness bug affecting any notebook step with boolean or None parameters. The hyphen-in-parameter-keys issue (Design #2) is a latent runtime error. The missing output_path validation (Maintainability #2) produces a poor user experience. The remaining findings are test coverage gaps and documentation staleness. After fixing the -y flag bug and adding paramet

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
+++ b/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
@@ -1097,6 +1097,7 @@ class NotebookStepConfig(BaseStepConfig):
             "papermill",
             shlex.quote(str(self.path)),
             shlex.quote(str(self.output_path)),
+            "-k python3",
         ]
         for key in sorted(self.parameters):
             value = self.parameters[key]
--- a/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
+++ b/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
@@ -324,6 +324,7 @@ class BaseStepConfig(ABC):
         to subclass-specific ``_build_from_dict`` for construction.
 
         Parameters
+        ----------
         data
             Dictionary from workflow YAML.
         output_directory
--- a/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
+++ b/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
@@ -974,6 +974,58 @@ class PythonStepConfig(BaseStepConfig):
 
 
 @dataclass
+class NotebookStepConfig(BaseStepConfig):
+    """Configuration for a notebook-based workflow step.
+
+    Currently routes to ``papermill`` under the hood, but exposes a
+    notebook-agnostic schema so the executor can be swapped later.
+    """
+
+    _SUPPORTED_ARGS: ClassVar[set[str]] = {"path", "parameters", "output_path", "cwd"}
+    _SCALAR_TYPES: ClassVar[tuple[type, ...]] = (str, int, float, bool)
+
+    name: str
+    """Unique name for this step within the workflow."""
+    resources: ResourceConfig
+    """Compute resources for this step."""
+    output_directory: Path
+    """Output directory for this step. Inherited from the workflow's output_directory."""
+    path: Path
+    """Path to the input notebook (.ipynb). Both relative and absolute paths are
+    accepted; resolved to absolute when serialized."""
+    output_path: Path
+    """Where to write the executed notebook (.ipynb). Both relative and absolute
+    paths are accepted."""
+    environment: str | None = None
+    """Optional environment name to use for this step."""
+    parameters: dict[str, Any] = field(default_factory=dict)
+    """Notebook parameters injected as cell-level variables. Scalar values only."""
+    cwd: Path | None = None
+    """Optional working directory for notebook execution. Both relative and absolute
+    paths are accepted."""
+
+    def _validate(self) -> None:
+        pass
+
+    def _build_command(self) -> str:
+        raise NotImplementedError
+
+    def to_dict(self) -> dict[str, Any]:
+        raise NotImplementedError
+
+    @classmethod
+    def _build_from_dict(
+        cls,
+        data: dict[str, Any],
+        output_directory: Path,
+        *,
+        project: str,
+        queue: str,
+    ) -> NotebookStepConfig:
+        raise NotImplementedError
+
+
+@dataclass
 class WorkflowConfig:
     """Parsed and validated workflow configuration."""
 
@@ -982,6 +1034,7 @@ class WorkflowConfig:
         "simulation": SimulationStepConfig,
         "pytest": PytestStepConfig,
         "python": PythonStepConfig,
+        "notebook": NotebookStepConfig,
     }
 
     name: str
--- a/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
+++ b/src/vivarium_cluster_tools/psimulate/workflow_config/config.py
@@ -47,6 +47,57 @@ DEFAULT_BACKUP_FREQ_SECONDS = 30.0 * 60.0
 """Default backup frequency in seconds (30 minutes), matching ``psimulate run``."""
 
 
+_SCALAR_TYPES: tuple[type, ...] = (str, int, float, bool)
+"""Scalar value types accepted in step args (e.g. notebook ``parameters``,
+python ``keyword_args`` / ``positional_args``)."""
+
+_IDENTIFIER_RE = re.compile(r"^[a-zA-Z0-9_-]+$")
+"""Pattern for keys in scalar-dict step args (``keyword_args``, ``parameters``)."""
+
+
+def _validate_scalar_dict(
+    configuration: object,
+    *,
+    field_name: str,
+    step_name: str,
+    allow_none_values: bool = True,
+) -> None:
+    """Validate ``configuration`` is a dict of identifier-keyed scalar values.
+
+    Parameters
+    ----------
+    configuration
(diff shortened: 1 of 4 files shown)


src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -47,6 +47,85 @@
...
 
+_SCALAR_TYPES: tuple[type, ...] = (str, int, float, bool)
+"""Scalar value types accepted in step args (e.g. notebook ``parameters``,
+python ``keyword_args`` / ``positional_args``)."""
+
+_IDENTIFIER_RE = re.compile(r"^[a-zA-Z0-9_-]+$")
+"""Pattern for keys in scalar-dict step args (``keyword_args``, ``parameters``)."""
+
+
+def _check_scalar(
```
albrja, line 58: More DRY

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -973,6 +1024,178 @@ def _build_from_dict(
...
+        if not str(self.path).endswith(".ipynb"):
+            raise ValueError(
+                f"Step '{self.name}': 'path' must end with .ipynb, " f"got {self.path!r}."
+            )
+        if not str(self.output_path).endswith(".ipynb"):
+            raise ValueError(
+                f"Step '{self.name}': 'output_path' must end with .ipynb, "
+                f"got {self.output_path!r}."
+            )
+        _validate_scalar_dict(
```
albrja, line 1102: Similar to the python step, we allow certain arguments required by papermill that we could change at a later date if we switched from papermill to a different library. Since papermill allows users to tag a notebook cell with the parameters tag, we cannot validate all arguments a user may pass in, so we do the same validation here that it must be a valid python variable name and value.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -47,6 +47,85 @@
...
+    if not isinstance(configuration, dict):
+        raise ValueError(
+            f"Step '{step_name}': '{field_name}' must be a dict, "
+            f"got {type(configuration).__name__}."
+        )
+    for key, value in configuration.items():
+        if not isinstance(key, str) or not _IDENTIFIER_RE.match(key):
+            raise ValueError(
+                f"Step '{step_name}': {field_name} key {key!r} is not a valid "
+                "identifier. Keys must match [a-zA-Z0-9_-]+."
```
rmudambi: Many users will not be fluent in regex. Say something like it must be alphanumeric or with dashes or underscores or that they must be valid Python identifiers.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -973,6 +1024,178 @@ def _build_from_dict(
...
+        _validate_scalar_dict(
+            self.parameters,
+            field_name="parameters",
+            step_name=self.name,
+        )
+        for key in self.parameters:
+            if not self._PYTHON_IDENTIFIER_RE.match(key):
+                raise ValueError(
+                    f"Step '{self.name}': parameter key {key!r} is not a valid "
+                    "Python identifier. Papermill requires parameter names that "
```
rmudambi: Don't reference Papermill in the error message.

rmudambi (approved):

patricktnast (approved):

PR #331, opened 2026-05-13
Support specifying archive node on cluster

albrja:
### Implement feature to specify whether a Jobmon task requires a archive cluster node
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6962

### Changes and notes
-support configuration for required archive node on slurm cluster
-the purpose of this is to be able to read data stored in certain places on the filesystem that non-archive nodes do not have access to

### Code reviwer
PR Review: Add requires_archive_node resource option for SLURM archive node constraint
Summary
This PR introduces a requires_archive_node: bool field to NativeSpecification (the low-level SLURM specification) and ResourceConfig (the YAML-facing workflow configuration). When set to True, tasks are constrained to SLURM nodes tagged with the archive feature. The implementation correctly builds SLURM --constraint expressions that AND-join the archive requirement with any hardware OR-group (parenthesizing appropriately for operator precedence). The feature is exposed only through workflow YAML configuration — the psimulate run/restart/expand CLI commands continue to use the False default.

Design
Boolean is the correct representation. "archive" is a SLURM feature tag, semantically distinct from hardware model names. Keeping it separate from hardware: list[str] avoids conflating two SLURM concepts and doesn't break validate_hardware(). No change needed.

Constraint algebra is correct. SLURM's & binds tighter than |, so r650|r650v2&archive would parse incorrectly. The interface.py:77 method correctly parenthesizes the hardware OR-group only when multiple hardware entries are combined with archive. The test matrix covers all quadrants.

Missing type validation on requires_archive_node from YAML input. At config.py:200-201, the value from data["requires_archive_node"] is passed through without type checking. If a user writes requires_archive_node: "true" (quoted string in YAML), it becomes a truthy string that behaves like True in boolean contexts but isn't actually a bool. Other fields (project, runtime, hardware) are validated in __post_init__, but this one is not. A string like "no" would be truthy and silently produce the wrong behavior.

Suggestion: Add to __post_init__:

if not isinstance(self.requires_archive_node, bool):    raise TypeError(        f"'requires_archive_node' must be a bool, "        f"got {type(self.requires_archive_node).__name__}: "        f"{self.requires_archive_node!r}."    )
Maintainability
Magic string "archive" in interface.py:84. The SLURM feature tag appears as a bare string literal. Extracting to a module-level constant (e.g., _SLURM_ARCHIVE_FEATURE = "archive") would improve discoverability and guard against typos.

Docstring placement on NamedTuple fields. The trailing docstrings for cores and requires_archive_node are not attached as __doc__ attributes at runtime (unlike dataclass fields). This is consistent with the existing pattern in the file, so no action needed, but worth noting if future Sphinx output seems incomplete.

DRY
Pre-existing: Four near-identical NativeSpecification(...) constructions in cli.py:170. The run, restart, expand, and load_test commands each construct the spec with the same 6 keyword arguments differing only in job_name. If requires_archive_node ever needs CLI exposure, all 4 must be updated. Extracting a helper would eliminate this risk. This is pre-existing, not introduced by the PR.
Tests
Missing round-trip test (from_dict → to_dict). The two directions are tested independently but not together. A round-trip test for both True and False values would catch regressions where one direction silently drops or flips the flag.

Missing YAML integration test. No test exercises requires_archive_node flowing through the full WorkflowConfig.from_yaml_with_cli_overrides pipeline. A test using make_step_dict(resources={"memory_gb": 4, "requires_archive_node": True}) through write_workflow_yaml → from_yaml_with_cli_overrides would verify YAML parsing preserves the value.

to_native_specification only tested for True; no assertion that False propagates. If the method accidentally hardcoded True, existing tests wouldn't catch it. Adding a one-line assertion for the default case is cheap insurance.

Documentation
CHANGELOG not updated. The 3.2.0 entry in CHANGELOG.rst:3 doesn't mention requires_archive_node. This is a user-facing configuration addition.

Suggestion: Add:

- Add ``requires_archive_node`` resource option for constraining workflow steps to SLURM archive nodes
ResourceConfig.requires_archive_node docstring doesn't explain interaction with hardware. Since this is the user-facing class, users writing YAML should understand that both fields combine into a single --constraint expression (e.g., (r650|r650v2)&archive) rather than being separate SLURM flags.

Suggestion: Append to the docstring at config.py:146:

When combined with ``hardware``, the archive requirement is AND-joinedwith the hardware constraint, e.g. ``(r650|r650v2)&archive``.
No user-facing documentation for the workflow YAML schema. The docs have no page describing the psimulate workflow command or its YAML format. The requires_archive_node option (and indeed the entire workflow feature) is undiscoverable without reading source. This is a broader gap that pre-dates this PR.

Functionality
to_dict() serialization is asymmetric but correct. When requires_archive_node is False, to_dict() omits the key entirely. When from_dict() receives a dict without the key, it defaults to False. This round-trips correctly. No issue here.

CLI callers are safe. All four NativeSpecification(...) calls in cli.py omit requires_archive_node, relying on the False default. Since the NamedTuple default is False, this is correct — CLI users don't get archive-pinned unintentionally.

No edge case issue with hardware=[] + requires_archive_node=True. _build_constraint() handles this by skipping the hardware branch (empty list is falsy) and only appending "archive". The test at id "archive-only" covers this case.

Minor Nits
The _build_constraint docstring at interface.py:77 could note it returns None when neither hardware nor archive is specified, to match the public to_jobmon_spec notes about "neither: the key is omitted."
Overall
This is a clean, well-scoped PR with correct SLURM constraint algebra and good test coverage. The three actionable items are:

Priority	Item
Medium	Add isinstance(self.requires_archive_node, bool) type validation in ResourceConfig.__post_init__
Low	Add CHANGELOG entry
Low	Document the hardware + requires_archive_node interaction in the ResourceConfig docstring

### DIFF
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -25,6 +25,8 @@ class NativeSpecification(NamedTuple):
     hardware: list[str]
     cores: int = 1
     """Number of CPU cores to request from SLURM. Default is 1."""
+    requires_archive_node: bool = False
+    """Whether the task must land on a node tagged with the SLURM ``archive`` feature."""
 
     def to_jobmon_spec(self, worker_logging_root: Path) -> dict[str, Any]:
         """Build the Jobmon compute resources dict from this NativeSpecification.
@@ -42,8 +44,16 @@ class NativeSpecification(NamedTuple):
         -----
         * ``memory`` is passed in **GB** because the Jobmon SLURM plugin performs
           its own GB → MB conversion internally.
-        * ``constraints`` is a pipe-separated string of SLURM feature names
-          (e.g. ``"r650|r650v2"``), included only when hardware is requested.
+        * ``constraints`` is a SLURM ``--constraint`` expression built from
+          ``hardware`` and ``requires_archive_node``:
+
+          - hardware only: pipe-joined (OR), e.g. ``"r650|r650v2"``.
+          - archive only: ``"archive"``.
+          - both: AND-joined, e.g. ``"r650&archive"`` or
+            ``"(r650|r650v2)&archive"`` (the hardware group is
+            parenthesized when it has more than one entry).
+          - neither: the key is omitted.
+
         * ``standard_output`` and ``standard_error`` route SLURM stdout/stderr
           to the cluster logs directory. The Jobmon SLURM plugin appends the
           task name and SLURM job ID to these paths automatically.
@@ -57,9 +67,24 @@ class NativeSpecification(NamedTuple):
             "stdout": str(worker_logging_root),
             "stderr": str(worker_logging_root),
         }
-        if self.hardware:
-            resources["constraints"] = "|".join(self.hardware)
+        constraint = self._build_constraint()
+        if constraint is not None:
+            resources["constraints"] = constraint
         return resources
+
+    def _build_constraint(self) -> str | None:
+        """Build the SLURM ``--constraint`` expression, or ``None`` if unconstrained."""
+        parts: list[str] = []
+        if self.hardware:
+            hw = "|".join(self.hardware)
+            if len(self.hardware) > 1 and self.requires_archive_node:
+                hw = f"({hw})"
+            parts.append(hw)
+        if self.requires_archive_node:
+            parts.append("archive")
+        if not parts:
+            return None
+        return "&".join(parts)
 
     @staticmethod
     def _runtime_to_seconds(runtime_str: str) -> int:
--- a/src/vivarium_cluster_tools/psimulate/cluster/interface.py
+++ b/src/vivarium_cluster_tools/psimulate/cluster/interface.py
@@ -45,14 +45,11 @@ class NativeSpecification(NamedTuple):
         * ``memory`` is passed in **GB** because the Jobmon SLURM plugin performs
           its own GB → MB conversion internally.
         * ``constraints`` is a SLURM ``--constraint`` expression built from
-          ``hardware`` and ``requires_archive_node``:
-
-          - hardware only: pipe-joined (OR), e.g. ``"r650|r650v2"``.
-          - archive only: ``"archive"``.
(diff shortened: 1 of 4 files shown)


hussain-jafari (approved):

stevebachmeier (approved):

src/vivarium_cluster_tools/psimulate/cluster/interface.py
```diff
@@ -57,10 +67,25 @@ def to_jobmon_spec(self, worker_logging_root: Path) -> dict[str, Any]:
...
+            resources["constraints"] = constraint
         return resources
 
+    def _build_constraint(self) -> str | None:
+        """Build the SLURM ``--constraint`` expression, or ``None`` if unconstrained."""
+        parts: list[str] = []
+        if self.hardware:
+            hw = "|".join(self.hardware)
+            if len(self.hardware) > 1 and self.requires_archive_node:
+                hw = f"({hw})"
```
rmudambi: Can we just always surround this in parentheses?

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -142,6 +142,13 @@ class ResourceConfig:
     """Number of CPU cores to request. Default is 1."""
     hardware: list[str] | None = None
     """Optional list of hardware types to target (e.g. ``["r650", "r650v2"]``)."""
+    requires_archive_node: bool = False
+    """Whether the step requires landing on an archive node.
+
+    When ``False`` (default), tasks may land on any node — archive or
+    non-archive. When ``True``, tasks are constrained to archive nodes
+    only (those tagged with the SLURM ``archive`` feature).
+    """
```
rmudambi: This should only be the first line.

rmudambi:
```suggestion
    """Whether to enforce landing on an archive node"""
```

rmudambi (approved):

PR #332, opened 2026-05-14
Support multiple parallel simulation steps

albrja:
### Fix configuration bug to support multiple simulation steps
- *Category*: 
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6962

### Changes and notes
-support multiple simulation steps


tests/psimulate/workflow_config/test_config.py
```diff
@@ -699,6 +699,7 @@ def test_get_tasks_wires_arguments(
                 worker_logging_root=Path("/out/logs"),
                 native_specification=config.native_specification,
                 env="test_env",
+                template_name="psimulate_sim_step",
```
stevebachmeier, line 710: Seems like you should actually add a test that shows this is truly supported

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -63,6 +64,10 @@ def get_task_list(
     env
         Optional conda environment name. When provided, the worker command
         is wrapped with ``conda run --no-capture-output -n <env>``.
+    template_name
+        Name to register the Jobmon ``TaskTemplate`` under. Must be unique
+        per Tool/Workflow; callers that build multiple simulation step
```
stevebachmeier, line 107: You say you need a distinct value per group but don't actually assert anything. What will happen?

stevebachmeier (commented):
I've lost the thread a bit on this but it seems like there's at least some tests that need to be added.

tests/psimulate/workflow_config/test_config.py
albrja: I am adding a test but it is rather difficult to write a test for this use case because we are really testing against the Jobmon API. The summary of the problem is, we enforce unique step names because a Jobmon workflow needs task template names to be unique and we pass the step name into each task as the template name. However, do to the preexisting code to run parallel simulations, we had hardcoded "psimulate" as the template name. This update is adding a argument to get_task_list for parallel simulations that defaults to "psimulate" but we overwrite it in the simulation step.

I have added a test but all of our tests that need to test the Jobmon API have to mock out quite a few things so it is hard to determine how good/useful the test actually is.

rmudambi: What if we had a weekly end-to-end test that actually runs a very small workflow on Jobmon?

albrja: Added test.

stevebachmeier (approved):

albrja:
A lot of environment handling was added to this step because of failures on the weekly e2e test. It raised some bugs that are Jenkins specific but fixing them means we can have e2e tests. Summary of the updates regarding handling the environment and moving away from payload wrapping "conda run ..."

Replaced the worker command's conda run -n <env> wrapper with PATH=<env_prefix>/bin:$PATH across all step types, resolving the env name to its absolute prefix once on the runner so workers no longer need conda on their PATH — fixing failures on Jenkins where the service user's conda is installed on host-local storage that compute nodes can't see

hussain-jafari (approved):

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -25,6 +28,45 @@
...
+    Uses ``CONDA_PREFIX`` directly when *env* matches the active env. For
+    other env names, queries ``conda env list --json`` via ``CONDA_EXE``
+    to find the matching prefix.
+
+    Worker shells on cluster compute nodes cannot always invoke ``conda``
+    (the conda binary itself may be installed on host-local storage), but
+    they can always invoke ``<env_prefix>/bin/<binary>`` because the env
+    lives on a shared filesystem.  Resolving the env to its prefix on the
+    runner lets us bake an env-local PATH into the worker command and
+    avoid any worker-side conda dependency.
```
rmudambi: Can remove this?

rmudambi (approved):

src/vivarium_cluster_tools/psimulate/runner.py
```diff
@@ -150,11 +150,10 @@ def workflow_main(
     )
 
     if wf_status != "D":
-        logger.warning(
```
albrja, line 153: Hey @patricktnast , do you remember why you had this as a warning instead of runtime error? I ran into issues debugging the new test in test_workflow_e2e Jobmon returns "D" or "F" and doesn't raise an exception or error. In other words: Jobmon "returned the failed workflow status" only inside the Python process — it never propagated out to the shell return code that the e2e test (and Jenkins) actually check.

patricktnast: I probably just didn't think that hard about it. i think raising is fine

patricktnast (approved):

PR #333, opened 2026-05-15
Albrja/mic 7036/workflow api

albrja:
### Add API for to support downstream library code to get instances of steps
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7036

### Changes and notes
-adds API for step classes 

### Code reviewer
PR Review: Albrja/mic 7036/workflow api ([#333](https://github.com/ihmeuw/vivarium_cluster_tools/pull/333))
Branches confirmed
Feature (head): albrja/mic-7036/workflow-api
Target (base): albrja/support-multiple-psims (stacked on another in-flight PR, not main)
Diff: 5 files, +563 / −17. Five commits.
Summary
Adds [src/vivarium_cluster_tools/psimulate/workflow_config/interface.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py) (NEW, 335 lines) — a Python API of six factory functions (get_command_step, get_simulation_step, get_pytest_step, get_python_step, get_notebook_step, and resolve_step_env_prefix) intended for downstream library callers who'd rather construct workflow steps programmatically than via YAML. Refactors [builder.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py) to delegate env-prefix resolution to the new helper (the env precedence logic was lifted cleanly out of WorkflowBuilder.build, not duplicated). Adds [tests/psimulate/workflow_config/test_interface.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py) (NEW, 221 lines) and a one-line patch-path update in [test_builder.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_builder.py). Also ignores docs/plans/ for local agent plan output.

Design
Most of the factory functions are isomorphic to their dataclass constructors and earn very little. (flagged by design, dry, maintainability)

[interface.py:31-65, 68-127, 130-177, 239-295](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L31-L295) — four of the five get_* functions are pure keyword passthroughs to the corresponding @dataclass that is already keyword-only with sensible defaults. The dataclasses don't hide any "scary" fields (native_specification is set in __post_init__, _SUPPORTED_ARGS is a ClassVar — neither is a constructor arg).
Costs: every new field on a step class must now be added in two places (signature + docstring + forwarding line), plus a third place in the test file. The wrappers also widen typing for path (string-only in PythonStepConfig, see §17 below) and quietly redefine defaults inline (see §5).
Suggestion: re-export the classes from interface.py (from .config import CommandStepConfig as CommandStep, …). Keep get_python_step only if you don't take §2 below, since it's the one factory that does non-trivial work. The PR motivation ("downstream callers want to build workflows from Python") is met by exporting the dataclasses directly.
PythonStepConfig.args as an untyped dict[str, Any] is the real design smell — the factory papers over it. (flagged by design, maintainability)

[config.py:905-1051](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L905-L1051) packs path / positional_args / keyword_args into one untyped dict, then every consumer does self.args["path"] / self.args.get(...). Sibling steps (Pytest, Notebook, Simulation) all promoted their args to first-class fields. PythonStepConfig is the odd one out.
[interface.py:225-235](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L225-L235) accepts the fields as separate kwargs and re-bundles them — which proves the right shape is already known. The bundling throws away types.
Suggestion: promote path: str, positional_args: list = field(default_factory=list), keyword_args: dict = field(default_factory=dict) to fields on PythonStepConfig; have to_dict / _build_from_dict translate at the YAML boundary. After that, get_python_step becomes a trivial wrapper like the others and can be deleted along with them (or kept for parity if you keep wrappers).
resolve_step_env_prefix is in the wrong module. (flagged by design, maintainability, dry)

The module docstring at [interface.py:1-10](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L1-L10) advertises "Python API for building workflow step configurations programmatically." resolve_step_env_prefix ([interface.py:298-335](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L298-L335)) is a build-time runtime helper, called only by [builder.py:76](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L76). The test file [test_builder.py:71](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_builder.py#L71) now has to patch …workflow_config.interface.resolve_env_prefix — patching one module's symbol through a sibling's namespace is a structural smell.
Suggestion: prefer making it a method on BaseStepConfig — step.resolve_env_prefix(default_environment=…) reads cleanly and puts the precedence chain on the step itself. Alternative: move it into builder.py as a private helper (it has one non-test caller).
PytestStepConfig.path: str | list[str] | None triple-union is exposed by the new public API. (flagged by design)

[config.py:830-836, 838-852, 891-896](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L830-L896) — three sites that branch on isinstance(self.path, list). Pre-existing, but the PR is the moment to normalize: coerce to list[str] in __post_init__, drop the triple-branching.
Default values are duplicated between factory signatures and dataclass fields. (flagged by maintainability, design)

[interface.py:77-78](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L77-L78) — backup_freq: float | None = DEFAULT_BACKUP_FREQ_SECONDS and sim_verbosity: int = 0 repeat the dataclass defaults at [config.py:622, 624](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L622-L624). SimulationStepConfig.to_dict ([config.py:728-731](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L728-L731)) treats those exact values as sentinels-to-omit. If DEFAULT_BACKUP_FREQ_SECONDS ever changes, a workflow YAML written by the old default will silently fail to round-trip.
Suggestion: use None as the API-layer sentinel and only forward kwargs that were explicitly set, e.g. if backup_freq is not None: kwargs["backup_freq"] = backup_freq. The dataclass remains the single source of defaults. This matches the pattern already used for positional_args/keyword_args in get_python_step.
Maintainability
resolve_step_env_prefix error message is now context-incoherent for programmatic callers. (flagged by maintainability)

[interface.py:330-334](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L330-L334) — "Set 'environment' on the step, 'default_environment' on the workflow, or activate a conda environment before running." When called via WorkflowBuilder.build this matches the YAML/CLI surface. When called by a Python user who hand-built a step config, "the workflow" doesn't yet exist in their call. Rephrase in terms of the function's own parameters: "Pass environment on the step, pass default_environment, or activate a non-base conda environment."
Asymmetric path handling between get_python_step and get_notebook_step. (flagged by maintainability)

get_python_step accepts path: str and silently bundles it into args. get_notebook_step accepts path: Path as a first-class kwarg. Two factories that look parallel behave differently — and the asymmetry surfaces the leaky PythonStepConfig.args representation. Resolved naturally by §2 above.
Type widening to Any loses contract that the dataclass enforces. (flagged by maintainability)

[interface.py:187-188](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L187-L188) — positional_args: list[Any] | None, keyword_args: dict[str, Any] | None. The whole point of a Python API over YAML is to give callers IDE/type-checker signal — but the dataclass validator restricts both to _SCALAR_TYPES = (str, int, float, bool) ([config.py:50](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L50)). Expose a Scalar type alias from config.py and annotate as list[Scalar] | None / dict[str, Scalar | None] | None. Same for parameters in get_notebook_step ([line 247](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L247)).
Pre-existing magic string "base". (flagged by maintainability)

[interface.py:329](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L329) — the meaning of "base" (conda installer env we refuse to run in) is not obvious. A one-line constant or comment when the function moved would have been the moment to capture this. Pre-existing; low priority.
DRY
The five happy-path API tests are structurally identical. (flagged by dry, tests)

[test_interface.py:33-122](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py#L33-L122) — each test does: build via API → isinstance → field-passthrough assertions. The validation test at [lines 125-169](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py#L125-L169) already shows the parametrize pattern is viable. Collapse them. Worth doing only if the factories stay (§1).
get_python_step re-encodes PythonStepConfig._SUPPORTED_ARGS inline. (flagged by dry)

[interface.py:225-235](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L225-L235) hard-codes the {"path", "positional_args", "keyword_args"} key set already declared at [config.py:946](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L946). A fourth supported arg added to the dataclass leaves the factory silently out of sync. Either drive the dict construction from _SUPPORTED_ARGS, or refactor per §2.
Re-import of resolve_env_prefix through interface.py is not duplication. (verified dry non-finding)

resolve_step_env_prefix is a legitimate wrapper (adds precedence + base check); the rename and the indirection are meaningful. Leave it.
Tests
Defaults are not exercised on any factory. (flagged by tests)

[test_interface.py:48-122](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py#L48-L122) — every happy-path test overrides every default. A refactor that, say, defaults artifact_path to Path(".") would not be caught. Add one defaults-only test per factory (backup_freq == DEFAULT_BACKUP_FREQ_SECONDS, sim_verbosity == 0, runslow is False, environment is None, parameters == {}, etc.).
get_python_step conditional branches at [interface.py:225-229](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L225-L229) are not fully covered. (flagged by tests)

The happy-path test hits both is not None branches simultaneously. There's no test for the omit case nor positional-only / keyword-only cases. The "omit when not provided" contract is the wrapper's only non-trivial public observable.
get_notebook_step(parameters=None) → {} normalization untested. (flagged by tests)

[interface.py:293](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L293) — only the dict-passed path is tested. If the wrapper were refactored to forward None, _validate_scalar_dict(None, …) would throw a confusing AttributeError.
get_pytest_step.path: list[str] form and k-only form untested. (flagged by tests)

The valid_pytest_paths fixture at [conftest.py:26-33](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/conftest.py#L26-L33) exists for this and is unused. The dataclass branches on isinstance(path, list) in three places — none are reached through the API surface in tests.
RuntimeError from resolve_env_prefix is promised by the docstring but untested. (flagged by tests)

[interface.py:325-326](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L325-L326) lists RuntimeError in Raises. The autouse fixture in TestResolveStepEnvPrefix always returns success. Add a case where resolve_env_prefix raises and assert the error propagates.
Python _validate path (.py extension check) is bypassed by the parametrized validation test. (flagged by tests)

[test_interface.py:148-149](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py#L148-L149) uses "/nonexistent/script.py" — that passes _validate and raises only from _validate_required_paths. Add a case with path="/tmp/not_a_script.txt" to exercise the .py suffix check at [config.py:968-972](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L968-L972).
Common-field forwarding (resources, output_directory) is not asserted in happy-path tests. (flagged by tests)

A regression that swapped resources and output_directory in any factory would not be caught. Easy fix once §10 collapses the tests.
Documentation
Missing Raises sections on every factory. (flagged by documentation)

get_command_step (ValueError), get_simulation_step (FileNotFoundError), get_pytest_step (ValueError, FileNotFoundError), get_python_step (ValueError, FileNotFoundError), get_notebook_step (ValueError, FileNotFoundError). The test suite at [test_interface.py:125-169](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_interface.py#L125-L169) treats these as part of the contract — they should be documented.
No Examples blocks on any factory. (flagged by documentation)

The dataclasses in config.py have YAML Examples blocks that are not useful for API callers. Add short Python Examples to at least get_python_step, get_notebook_step, get_simulation_step.
Missing Sphinx page for the new module. (flagged by documentation)

The pattern under docs/source/api_reference/psimulate/workflow_config/ is one .rst per submodule, included via :glob: *. A sibling interface.rst with .. automodule:: is needed or the user-facing API does not appear in built docs.
CHANGELOG entry too thin. (flagged by documentation)

The 3.2.0 stanza in CHANGELOG.rst lacks a bullet for the new Python API surface. Users won't learn about it from the changelog.
get_python_step.path annotation/doc asymmetry. (flagged by documentation, design)

[interface.py:185, 209-211](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L185-L211) — annotated str, but the prose "Both relative and absolute paths are accepted" is identical to other path: Path factories. PythonStepConfig._validate ([config.py:968](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L968)) actively rejects pathlib.Path. Tighten the doc or widen the type with a str() coercion inside the factory.
get_pytest_step.path annotation/doc asymmetry. (flagged by documentation)

Same problem: str | list[str] | None, with Path not accepted, while every other factory takes Path. Either widen or call out the difference. Resolved naturally by §4.
get_python_step.keyword_args key-rule documentation is inaccurate. (flagged by documentation)

[interface.py:217-219](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L217-L219) — docstring says "identifier-style keys" but _IDENTIFIER_RE = r"^[a-zA-Z0-9_-]+$" ([config.py:54](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L54)) permits dashes and leading digits — not Python identifiers. Tighten: "Keys must match [A-Za-z0-9_-]+."
get_python_step None→bare-flag rule is correct but counter-intuitive. (flagged by documentation)

Most CLI conventions treat None as "omit" — here None is treated identically to True. Add a one-line callout.
Module docstring is too sparse for a new public API. (flagged by documentation)

[interface.py:1-10](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L1-L10) — two sentences. Should at minimum list the factories and link to WorkflowConfig/BaseStepConfig.
Functionality
Cross-package dependency direction reversal. Before this PR, builder.py imported resolve_env_prefix directly from jobmon_config.workflow. Now builder.py imports from a sibling module interface.py, which in turn imports from jobmon_config.workflow. Functionally fine, but the new graph is: builder → interface → jobmon_config, where interface is supposed to be the user-facing layer. Re-locating resolve_step_env_prefix per §3 fixes this.

Behaviorally equivalent env-resolution refactor — verified. The old inline block in WorkflowBuilder.build and the new resolve_step_env_prefix produce the same result on the same inputs (same precedence chain, same error message text, same resolve_env_prefix call). The only observable difference is the location of the raise in the call stack. No behavior regression.

No edge-case regressions detected. Empty-string environment continues to fall through ("" is falsy). default_environment=None is equivalent to omission. step.environment == "base" is rejected as before. is_resume / build_timestamp are passed through unchanged.

Minor Nits
.gitignore lacks trailing newline. [.gitignore:124](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/.gitignore#L124) — the new docs/plans/ line is added on the no-newline-at-end-of-file line. Add a trailing newline.

Trailing space / quoting in SimulationStepConfig._build_command error message. [config.py:649](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L649) — "SimulationStepConfig does not use _build_command. " "Use get_tasks() instead." is two adjacent string literals; harmless, but unintentional-looking. Pre-existing.

Pre-existing TODO in builder. [builder.py:56](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L56) — # TODO: MIC-6997 - encapsulate Jobmon UI in one place. This PR is a step in that direction. Either reference the new helper or leave a note in the ticket. Pre-existing; no action needed.

get_python_step.path is required but has no default and is in a *, (keyword-only) section — small UX paper cut. If you keep the wrapper, putting path: str earlier in the param list won't help since everything is keyword-only. Skip.

Overall
This PR introduces a clean, well-tested abstraction layer that mostly forwards to dataclasses that are already keyword-only Python APIs. The biggest open questions are structural:

Does this layer earn its keep? Four of the five factories add zero behavior. The "Python API" goal is satisfied by exporting the existing dataclasses. The wrapper layer creates parallel signatures, parallel docstrings, parallel default values, and parallel tests — every one a place to drift.
PythonStepConfig.args is the only step config that uses an untyped dict. The factory's bundling logic exists because of that representation. Fixing the dataclass (promoting path/positional_args/keyword_args to typed fields) would simplify both layers.
resolve_step_env_prefix does not belong in interface.py. It's a build-time helper. Make it a method on BaseStepConfig or move it back into builder.py.
If the wrappers stay, the highest-value fixes are: (a) document Raises sections and add Examples; (b) wire up interface.rst and a CHANGELOG bullet so the API is discoverable; (c) close the test gaps around defaults, omit-when-unset semantics, and the unexercised pytest list[str] / k-only / RuntimeError-propagation paths; (d) replace the inline default duplication (DEFAULT_BACKUP_FREQ_SECONDS, sim_verbosity=0) with None sentinels so the dataclass remains the single source of truth.

Note for the merge: this PR targets albrja/support-multiple-psims, not main. Worth confirming that base is the intended one (and that the base's changes have already been considered) before final review on main.


src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -104,40 +87,10 @@ def build(self, workflow_args: str) -> Workflow:
...
-    def _is_resume(self) -> bool:
-        """Check whether this is a resumed workflow build.
-
-        Returns True if the build timestamp file already exists in the
-        output directory, indicating a previous build has run.
-        """
-        timestamp_path = self.config.output_directory / BUILD_TIMESTAMP_FILENAME
-        return timestamp_path.exists()
-
     def _get_or_create_build_timestamp(self) -> str:
```
rmudambi: This method now seems unnecessary indirection.

src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py
```diff
@@ -0,0 +1,177 @@
...
+
+_SCALAR_TYPES: tuple[type, ...] = (str, int, float, bool)
+"""Scalar value types accepted in step args (e.g. notebook ``parameters``,
+python ``keyword_args`` / ``positional_args``)."""
+
+_IDENTIFIER_RE = re.compile(r"^[a-zA-Z0-9_-]+$")
+"""Pattern for keys in scalar-dict step args (``keyword_args``, ``parameters``)."""
+
+
+def _get_or_create_build_timestamp(output_directory: Path) -> str:
```
rmudambi: You have several utility functions in this class that have _ prefixes, which is inappropriate.

src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py
```diff
@@ -0,0 +1,177 @@
...
+    if timestamp_path.exists():
+        return timestamp_path.read_text().strip()
+
+    build_timestamp = datetime.now().strftime("%Y_%m_%d_%H_%M_%S")
+    output_directory.mkdir(parents=True, exist_ok=True)
+    timestamp_path.write_text(build_timestamp)
+    return build_timestamp
+
+
+def is_build_resume(output_directory: Path) -> bool:
```
rmudambi: Why not call this `is_resume()`?

.gitignore
```diff
@@ -120,4 +120,5 @@ src/vivarium_cluster_tools/_version.py
 .github/copilot_instructions.md
 
 # Agent files
-.claude/
\ No newline at end of file
+.claude/
+docs/plans/
```
rmudambi: What is in this directory and why is it being gitignored?

rmudambi (approved):

rmudambi (approved):

patricktnast (approved):

PR #335, opened 2026-05-18
Albrja/refactor/mic 7115/wrap api functions

albrja:
### Refactor workflow yaml configuration path to loosely wrap API functions
- *Category*: Refactor
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7115

### Changes and notes
-refactor workflow config and workflow builder to loosely wrap API functions

### Code reviewer
PR Review: Refactor config and build to loosely wrap API functions (#335)
Branches confirmed: Feature albrja/refactor/mic-7115/wrap-api-functions → Target albrja/mic-7036/workflow-api (stacked PR — NOT main).

Note on working tree: Your local [src/vivarium_cluster_tools/psimulate/workflow_config/builder.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py) has uncommitted changes per git status. The documentation reviewer flagged that the on-disk file is missing STEP_TYPE_TO_API_FN. The review below is based on the PR head (origin/refs/pull/335/head), not your working tree.

Summary
This PR refactors WorkflowConfig and WorkflowBuilder so that the parsed config holds a new frozen dataclass, ParsedStep, instead of typed BaseStepConfig instances. Each concrete step config gains a to_api_kwargs() method that produces a kwargs dict matching the corresponding get_*_step_tasks interface function in the parent PR (MIC-7036). The builder then dispatches by step_type string via a new STEP_TYPE_TO_API_FN registry, applies workflow-level default_environment, and invokes the interface API function. After parsing, the intermediate BaseStepConfig instance is discarded; ParsedStep caches both the API kwargs (for build) and the YAML form (for round-trip serialization).

The change is sensibly motivated (config becomes a thin wrapper around the public programmatic API), but it introduces meaningful type-safety loss, three parallel step-type registries, and a couple of quiet behavioral changes.

Design
Type erasure on WorkflowConfig.steps — [config.py:1293](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1293). list[BaseStepConfig] → list[ParsedStep] with api_kwargs: dict[str, Any]. Downstream consumers now do string-keyed dict access (step.api_kwargs["name"], step.api_kwargs["resources"].memory_gb) instead of attribute access. mypy/IDE can't catch typos. Alternative: keep the BaseStepConfig instance on ParsedStep and derive kwargs lazily — preserves dispatch decoupling, drops yaml_dict redundancy, keeps types.
Three parallel step-type registries — STEP_TYPE_TO_API_FN in [builder.py:33-40](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L33-L40), ParsedStep._STEP_CLASS_TO_TYPE at [config.py:1234-1240](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1234-L1240), and WorkflowConfig.SUPPORTED_STEP_TYPES at [config.py:1274-1280](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1274-L1280) all key on the same five strings. Adding a new step type requires updating all three; drift yields silent KeyError or misleading errors. This is the highest-value design fix. Even reducing 3→2 by collapsing SUPPORTED_STEP_TYPES and _STEP_CLASS_TO_TYPE into one bidirectional source is worthwhile. Best: one canonical STEP_TYPES: dict[str, tuple[type[BaseStepConfig], Callable]] in either config.py or interface.py (mind the potential config ↔ interface import cycle).
to_api_kwargs lives on the wrong class — [config.py:428-437](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L428-L437) (abstract) plus 5 implementations. It's an adapter between configs and interface function signatures, consumed only by the builder, but it lives on BaseStepConfig. Couples config classes to interface signatures across two files. Consider free functions in interface.py (or a single registry that pairs config class → (api_fn, kwarg_extractor)), keeping the kwarg-key string list next to the function that consumes it.
frozen=True gives a false impression of immutability — [config.py:1223](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1223). Frozen only blocks attribute rebinding; the two dict fields (api_kwargs, yaml_dict) are deeply mutable. The builder's _resolve_environment comment ([builder.py:91-92](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L91-L92)) is the only thing preserving the no-mutation invariant. Consider types.MappingProxyType(...) snapshots, or drop frozen=True and document the contract.
_resolve_environment is a pure function masquerading as a method — [builder.py:87-96](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L87-L96). Reads only self.config.default_environment. Pull it out as a free function in utilities.py (signature (api_kwargs, default_environment) → dict); unit tests no longer need to construct a full WorkflowBuilder/WorkflowConfig pair.
_resolve_environment blindly injects "environment" — same site. Safe today because all five interface functions accept environment=, but the design hands the builder a generic dict[str, Any] and trusts every entry in STEP_TYPE_TO_API_FN to accept that kwarg. A future step type whose API has no environment would explode at runtime. Either (a) only inject when key already present in api_kwargs, or (b) document the contract on BaseStepConfig.to_api_kwargs.
Redundant snapshotting of api_kwargs and yaml_dict — [config.py:1247-1252](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1247-L1252). Both are projections of the same dataclass. Silent divergence is possible ("to_dict emits path, to_api_kwargs emits script_path"). If ParsedStep.config keeps the typed instance (finding #1 alternative), yaml_dict disappears.
(Pre-existing) O(n²) duplicate-name detection — [config.py:1454-1458](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1454-L1458). names.count(name) for name in names. Irrelevant at realistic sizes, but the rewrite is shorter: Counter(names)-based. Touching code anyway, so easy to fold in.
Maintainability
step.api_kwargs["name"] elevates "name" to an undeclared required key — [config.py:1454](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1454). ParsedStep advertises api_kwargs only as "kwargs ready to splat into the matching interface API function" — nothing says omitting "name" triggers a KeyError in __post_init__. Fix: promote name to a first-class field on ParsedStep (name: str), set via step.name in from_step_config, and the validation reverts to step.name. Also cleans up the test-suite step.api_kwargs["name"] idiom.
to_api_kwargs drift from API signatures is undetectable until runtime — [config.py:482-490, 677-689, 830-840, 999-1014, 1171-1184](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L482). A renamed or added param on a get_*_step_tasks function fails lazily — only when a workflow of that type is built, with a TypeError that doesn't point at the offending method. See Tests finding #1 for a one-shot drift-detection test using inspect.signature.bind_partial.
valid_yaml_types excludes "command" from the error message — [config.py:1318, 1356-1361](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1318). The intent is "specify a command step by writing command:, not type: command," but type: command is now silently accepted (see Functionality #1). A user who tries type: command will get a message saying that's not a valid type — when it actually is. Pick one: include "command" in the message, OR raise on explicit type: command.
mock_resolve_env_prefix patches a name that's still reachable, but the call chain has moved — [tests/psimulate/workflow_config/test_builder.py:51-61](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_builder.py#L51-L61). The patch hits utilities.resolve_env_prefix (the name imported into utilities). It works because utilities.resolve_step_env_prefix calls resolve_env_prefix as a module attribute. If a future refactor in utilities.py localizes the import or in interface.py calls utilities.resolve_env_prefix directly, the mock silently stops working and tests would write to the real filesystem. Suggestion: patch at the call site (interface.resolve_step_env_prefix) or document the assumption.
Build-timestamp comment describes behavior in a different module — [builder.py:71-73](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L71-L73). The builder no longer calls get_or_create_build_timestamp — the side-effect happens deep inside each get_*_step_tasks in interface.py. The spatial separation makes the ordering invariant hard to verify. Consider an explicit get_or_create_build_timestamp(...) call in the builder before the loop (matching the old code's clarity) and have interface functions accept a pre-computed timestamp.
ParsedStep.from_step_config is the only safe construction path, but the public 3-field constructor is not blocked — [config.py:1254-1266](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1254-L1266). Direct ParsedStep(step_type=..., api_kwargs=..., yaml_dict=...) calls produce an unvalidated instance. Add __post_init__ validation, or document the factory as the only entry point.
DRY
Per-class to_api_kwargs boilerplate (acceptable) — Every implementation opens with the same four keys (name, resources, output_directory, environment). The explicit form is the right call here: it documents the function-specific contract clearly, and dataclasses.asdict doesn't fit because of PythonStepConfig.args unpacking. Optional: _base_api_kwargs() helper on BaseStepConfig if a sixth step type lands.
Three parallel registries — Same finding as Design #2 above. Highest-value DRY consolidation: at minimum, collapse SUPPORTED_STEP_TYPES + _STEP_CLASS_TO_TYPE (config.py only) into one bidirectional source via _STEP_CLASS_TO_TYPE = {v: k for k, v in SUPPORTED_STEP_TYPES.items()}.
config.steps[i].api_kwargs["foo"] repetition in tests (acceptable as-is, but add ParsedStep.name) — Repeated across ~14 sites in [tests/psimulate/workflow_config/test_config.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py). Pull name out as a property on ParsedStep (also addresses Maintainability #1). Other accesses are read-friendly as explicit dict lookups.
Test fixture wrapping ParsedStep.from_step_config(CommandStepConfig(...)) — [tests/psimulate/workflow_config/test_builder.py:30-39, 97-106](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_builder.py#L30) and [tests/psimulate/test_runner.py:388-400](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/test_runner.py#L388). The PR nicely collapsed two adjacent sites into a comprehension. test_runner.py was not updated to share. Optional helper make_command_step(name, command); defensible to leave as 3 small fixtures.
Tests
No test asserts to_api_kwargs() keys match the API fn signature — All five to_api_kwargs are only exercised transitively (_parse_steps → fixtures) and assertions in [test_config.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py) only spot-check 1–2 keys. Add one parametrized test using inspect.signature(STEP_TYPE_TO_API_FN[step_type]).parameters to assert exactly set(api_kwargs) == sig - {"tool", "is_resume"}. This is the single highest-value test addition — it locks in the contract claimed by [config.py:430-436](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L430).
WorkflowBuilder.build dispatch is only tested for "command" — Both three_step_config and _make_single_step_config construct only CommandStepConfig steps. The other four entries in STEP_TYPE_TO_API_FN are exercised only via test_interface.py, never via the builder's registry dispatch. A typo in any non-command key in STEP_TYPE_TO_API_FN would not be caught. Add a parametrized test_build_dispatches_to_matching_api_fn that patches each registry entry to a mock and verifies the right mock is invoked.
_resolve_environment immutability is undocumented in tests — The docstring promises "Returns a new dict so the cached ParsedStep.api_kwargs is not mutated." There is no test for this. (See also Functionality #2 — the docstring is partially inaccurate.)
ParsedStep.from_step_config class-to-type round-trip is untested — _STEP_CLASS_TO_TYPE is the source of truth for the class→string mapping but is only ever exercised via YAML parsing. A programmatic test (the documented use case for from_step_config) would lock in current mappings.
"Unsupported type" error message is untested — The PR materially changes the message (excludes "command"), but grep "unsupported type" tests/ returns nothing. Add: (a) one test that type: not_a_type raises with the expected match, and (b) one test that the listed valid types do not include "command".
step.api_kwargs["..."] brittleness scattered across the suite — See DRY #3. Use a ParsedStep.name property for the most common access; leave other dict lookups explicit.
isinstance(step, …Config) → step.step_type == "…" — A real loss of strength (no longer catches a bug where _STEP_CLASS_TO_TYPE returns the wrong string), but the test in finding #4 above mitigates this if added.
Documentation
STEP_TYPE_TO_API_FN docstring — [builder.py:41](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L41). Adequate but should note the registry must stay in sync with WorkflowConfig.SUPPORTED_STEP_TYPES and ParsedStep._STEP_CLASS_TO_TYPE — the three-registry coupling is otherwise easy to forget.
WorkflowConfig.steps field docstring — [config.py:1293-1295](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1293-L1295). Accurate for the build path but elides that ParsedStep also carries the YAML form used by to_dict(). Mention both responsibilities.
Asymmetric to_api_kwargs docstrings — Four implementations have one-line docstrings; only PythonStepConfig.to_api_kwargs ([config.py:1002-1018](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1002-L1018)) gets the longer explanation (unpacking args). NotebookStepConfig.to_api_kwargs ([config.py:1174](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1174)) also has nontrivial behavior (parameters default of {} vs. interface fn's None | dict allowance) but no explanation.
CHANGELOG.rst not updated — [CHANGELOG.rst](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/CHANGELOG.rst) has an open 3.2.0 - TBD/TBD/TBD section. If the parent PR (MIC-7036) owns the changelog entry for the public get_*_step_tasks interface, this PR can skip the entry. Confirm with stack ordering. README is correctly not updated (no user-visible CLI changes).
Functionality
Quiet behavioral change: type: command is now accepted in YAML — In the parent branch (base-7036), SUPPORTED_STEP_TYPES did NOT contain "command", so type: command raised ValueError. In this PR, "command" is in SUPPORTED_STEP_TYPES ([config.py:1276](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1276)), so type: command parses silently as a command step. The error message in [config.py:1358-1361](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1358-L1361) still hides "command" from the valid-types list. Either accept this and update the message to include "command", or restore the old rejection. The PR description and changelog don't mention the change.
Misleading docstring on _resolve_environment — [builder.py:88-92](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L88-L92). Says "Returns a new dict so the cached ParsedStep.api_kwargs is not mutated." Only the second branch ([builder.py:96](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L96)) returns a new dict; the first branch returns api_kwargs unchanged — the same object as the cache. Currently safe because the caller immediately splats the result, but the comment overstates the guarantee. Either copy in both branches (return dict(api_kwargs)) or correct the docstring.
Possible silent fix of a pre-existing resume-detection bug — In the parent branch's [builder.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py), the order was:

build_timestamp = get_or_create_build_timestamp(self.config.output_directory)  # writes marker
resuming = is_resume(self.config.output_directory)  # always True!
Because is_resume checks for BUILD_TIMESTAMP_FILENAME ([utilities.py:63-66](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py#L63-L66)) which get_or_create_build_timestamp writes ([utilities.py:44-60](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py#L44-L60)), resuming was always True after the first line ran — including on fresh builds. This PR reorders so is_resume is checked before any side effect, which correctly returns False on fresh builds. If my reading is correct, this is a real (if previously latent) bug fix — call it out explicitly in the PR description and verify the downstream effect of is_resume=True on fresh builds was previously benign. If it wasn't benign, this PR fixes a subtle correctness issue.
STEP_TYPE_TO_API_FN[parsed_step.step_type] will raise KeyError for a programmatically-built ParsedStep with an invalid step_type — [builder.py:81-82](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/builder.py#L81-L82). YAML-built steps validate against SUPPORTED_STEP_TYPES, so unreachable from that path. But ParsedStep's public constructor has no validation. A __post_init__ check (assert step_type in _STEP_CLASS_TO_TYPE.values()) would fail loudly at construction.
Minor Nits
The valid_yaml_types variable name in [config.py:1318](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1318) is misleading — "command" IS valid in YAML now, just hidden. Rename to documented_yaml_types or non_default_step_types, or drop the filter entirely (see Maintainability #3).
Pre-existing O(n²) duplicate-name check at [config.py:1454-1458](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L1454-L1458) — easy Counter-based rewrite while touching this code.
Overall
The refactor's goal — letting the YAML config and programmatic API share one implementation path — is the right direction, and the ParsedStep abstraction does decouple the builder from step-class identity. But the realization trades typed BaseStepConfig instances for a dict[str, Any] whose schema is enforced only by convention, and it introduces a three-way registry coupling that the type system cannot enforce.

Highest-impact changes before merge:

Collapse the three step-type registries (Design #2 / DRY #2 / Maintainability #1). At minimum, derive _STEP_CLASS_TO_TYPE from SUPPORTED_STEP_TYPES so two of the three are guaranteed to agree.
Add the signature-drift test (Tests #1) — single parametrized test, locks in the to_api_kwargs ↔ get_*_step_tasks contract.
Resolve the type: command inconsistency (Functionality #1 / Maintainability #3) — pick one of "accept and document" or "reject" and align the error message.
Promote name to a ParsedStep field (Maintainability #1) — fixes the implicit-required-key footgun and cleans up tests.
Verify or document the resume-ordering fix (Functionality #3) — if it's a real fix, call it out in the PR description.
The remaining design-level concerns (type erasure, to_api_kwargs placement, frozen=True's false guarantee) are real but more invasive — defensible to defer if the team is comfortable with the convention-driven contract, but worth a follow-up issue.


src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -1143,18 +1220,61 @@ def _build_from_dict(
...
+
+
+_STEP_CLASS_TO_TYPE: dict[type[BaseStepConfig], str] = {
+    cls: step_type for step_type, cls in STEP_TYPES.items()
+}
+"""Derived inverse of :data:`STEP_TYPES` for fast ``step class → step_type`` lookup."""
+
+
+@dataclass(frozen=True)
+class ParsedStep:
```
albrja, line 158: This is basically just a hand off class that helps enforce contracts between the various steps and passing the data to the API functions.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -55,24 +52,18 @@ def build(self, workflow_args: str) -> Workflow:
             default_max_attempts=self.config.max_attempts,
         )
 
-        # Generate a stable build timestamp once per workflow build.
-        # On resume, reuse the timestamp from the previous build so that
-        # steps produce identical output paths.
-        build_timestamp = get_or_create_build_timestamp(self.config.output_directory)
+        # Resume must be checked before any step runs: the build-timestamp
+        # marker is what is_resume looks for, and the first interface API
+        # call will write that marker as a side-effect of running.
```
rmudambi: This comment seems superfluous

src/vivarium_cluster_tools/psimulate/workflow_config/interface.py
```diff
@@ -363,3 +364,14 @@ def get_notebook_step_tasks(
...
+
+STEP_TYPE_API_FNS: dict[str, Callable[..., list[Task]]] = {
+    "command": get_command_step_tasks,
+    "simulation": get_simulation_step_tasks,
+    "pytest": get_pytest_step_tasks,
+    "python": get_python_step_tasks,
+    "notebook": get_notebook_step_tasks,
+}
+"""Maps each YAML ``step_type`` to the API function that builds its tasks.
+Paired with :data:`vivarium_cluster_tools.psimulate.workflow_config.config.STEP_TYPES`."""
```
rmudambi: This seems better situated in builder.py

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -86,3 +77,14 @@ def build(self, workflow_args: str) -> Workflow:
...
+    def _resolve_environment(self, api_kwargs: dict[str, Any]) -> dict[str, Any]:
+        """Apply workflow-level ``default_environment`` to a step's kwargs.
+
+        Step-level ``environment`` wins; otherwise the workflow default is
+        substituted. Returns a new dict so the cached ``ParsedStep.api_kwargs``
+        is not mutated.
+        """
+        if api_kwargs.get("environment") is not None:
+            return api_kwargs
+        return {**api_kwargs, "environment": self.config.default_environment}
```
rmudambi: This is 3 lines. It doesn't seem worthwhile to split it out into a separate method.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -1143,18 +1220,61 @@ def _build_from_dict(
...
+    "command": CommandStepConfig,
+    "simulation": SimulationStepConfig,
+    "pytest": PytestStepConfig,
+    "python": PythonStepConfig,
+    "notebook": NotebookStepConfig,
+}
+"""Maps each YAML ``step_type`` string to its step-config class. Adding a new
+step type requires a matching entry in
+:data:`vivarium_cluster_tools.psimulate.workflow_config.interface.STEP_TYPE_API_FNS`"""
+
```
rmudambi: This could be inside `ParsedStep`

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -1210,46 +1342,45 @@ def _parse_steps(
...
+                    "Use the bare 'command' field instead (omit 'type')."
                 )
-            else:
-                # Default command-based step
-                step = CommandStepConfig.from_dict(
-                    step_dict,
-                    output_directory=output_directory,
-                    project=project,
-                    queue=queue,
+            step_type = explicit_type or "command"
```
rmudambi: I don't see why we should error if the user provides a command and `type == "command`. 
```python
            step_type = step_dict.get("type", "command")
            if step_type == "command" and "command" not in step_dict:
                raise ValueError(...)
```

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -1210,46 +1342,45 @@ def _parse_steps(
...
-            steps.append(step)
-        return steps
+            step_class = STEP_TYPES[step_type]
+            step = step_class.from_dict(
+                step_dict,
+                output_directory=output_directory,
+                project=project,
+                queue=queue,
+            )
+            parsed_steps.append(ParsedStep.from_step_config(step))
```
rmudambi: There's no reason why we should need to instantiate a StepConfig object here. We can just have a class-level function to generate a `ParsedStep`. Also, the strategy for creating a ParsedStep object is the same for all step types except the python step (look for the static list of parameters for the interface function in the config and add them to the dictionary). 

I guess you're using the instantiation to apply the validations, but that is not going to be easily maintainable. You should be more explicit about how validation is being done.

rmudambi (changes requested):

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
albrja: I mean it could but we need it before we actually create an instance so it would need to be a class var. It also is a bit weird to be to add a global mapping to a class.

src/vivarium_cluster_tools/psimulate/workflow_config/interface.py
```diff
@@ -68,6 +68,13 @@ def get_command_step_tasks(
...
         The Jobmon tasks produced by the step.
     """
+    CommandStepConfig.validate(
+        name=name,
+        resources=resources,
+        command=command,
+        output_directory=output_directory,
+        environment=environment,
+    )
     step = CommandStepConfig(
```
albrja: I couldn't entirely remove making instances since the instances are doing two things:
-Typed parameter bag for the runtime methods.
-Polymorphic dispatch for _build_command and get_tasks overrides (notably SimulationStepConfig).

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -51,18 +68,17 @@ def build(self, workflow_args: str) -> Workflow:
...
         all_tasks: list[Task] = []
 
         for parsed_step in self.config.steps:
             api_fn = STEP_TYPE_API_FNS[parsed_step.step_type]
-            kwargs = self._resolve_environment(parsed_step.api_kwargs)
+            # Step-level environment wins; otherwise the workflow default is
+            # substituted. Build a new dict so the cached api_kwargs is not mutated.
+            kwargs = parsed_step.api_kwargs
+            if kwargs.get("environment") is None:
+                kwargs = {**kwargs, "environment": self.config.default_environment}
```
rmudambi: Why would we care if this is mutated? It was created during the config parsing process, and will never be used again.

src/vivarium_cluster_tools/psimulate/workflow_config/config.py
```diff
@@ -173,13 +172,11 @@ def to_native_specification(self, job_name: str) -> NativeSpecification:
 class BaseStepConfig(ABC):
```
rmudambi: These aren't really configs anymore. `BaseStep` and similar below makes more sense to me.

src/vivarium_cluster_tools/psimulate/workflow_config/builder.py
```diff
@@ -9,29 +9,43 @@
...
+    "simulation": get_simulation_step_tasks,
+    "pytest": get_pytest_step_tasks,
+    "python": get_python_step_tasks,
+    "notebook": get_notebook_step_tasks,
+}
+"""Maps each YAML ``step_type`` to the API function that builds its tasks.
+Paired with :data:`vivarium_cluster_tools.psimulate.workflow_config.config.STEP_TYPES`."""
+
+
 class WorkflowBuilder:
```
rmudambi: This doesn't actually need to be a class. We create it exactly once and then use it to call `build()`. There is no persistent state we need to manage. It can just be a function that builds a Jobmon Workflow.

src/vivarium_cluster_tools/psimulate/workflow_config/interface.py
rmudambi: I don't think I agree with this statement. There is no state that is stored in a StepConfig object beyond that which is already in a ParsedStep. We could retain the classes if we think that is a useful way of organizing code specific to a give step type (i.e. validations and bespoke logic for creating the appropriate Jobmon tasks), but I don't see any reason we need to instantiate them. 

We currently have to instantiate them because you have put some of the validation logic inside the __post_init__ method, but that isn't really a great location for validation - it'd be better if that was in an explicit validation function.

src/vivarium_cluster_tools/psimulate/workflow_config/task_builders.py
```diff
@@ -0,0 +1,282 @@
...
+
+def _create_single_command_task(
+    tool: Tool,
+    *,
+    name: str,
+    resources: ResourceConfig,
+    output_directory: Path,
+    env_prefix: str,
+    command: str,
+) -> Task:
```
rmudambi: Whenever you call this, you're wrapping the return in a list. You might as well just have this method return the list instead.

src/vivarium_cluster_tools/psimulate/workflow_config/task_builders.py
```diff
@@ -0,0 +1,282 @@
...
+            name=name,
+            resources=resources,
+            output_directory=output_directory,
+            env_prefix=env_prefix,
+            command=" ".join(parts),
+        )
+    ]
+
+
+def _create_single_command_task(
```
rmudambi: Nit: I prefer `_get_single_command_task()`. When I see a method called `create_*()`, I instinctively think there is something being created and stored somewhere else. `get_*()` communicates more clearly that this method is returning something to the caller, and it's the caller's responsibility to do something intelligent with it.

src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py
```diff
@@ -67,22 +62,25 @@ def is_resume(output_directory: Path) -> bool:
 
 
 def resolve_step_env_prefix(
-    step: BaseStepConfig,
     *,
```
rmudambi, line 76: Remind me why we want to disallow positional arguments?

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
```diff
@@ -0,0 +1,186 @@
...
+    check_scalar,
+    validate_scalar_dict,
+)
+
+_PYTHON_IDENTIFIER_RE = re.compile(r"^[a-zA-Z_][a-zA-Z0-9_]*$")
+"""Notebook parameter keys must be valid Python identifiers because papermill
+injects them as variable assignments in a notebook cell."""
+
+
+def validate_common(name: str, resources: ResourceConfig) -> None:
```
rmudambi: Nit: this can be `_validate_common()`?

Same for the two methods below.

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
```diff
@@ -0,0 +1,186 @@
...
+    """Validation shared by every step type: name + resource resolution."""
+    if not name:
+        raise ValueError("Step 'name' is required.")
+    if not resources:
```
rmudambi: Is it possible for a `ResourceConfig` object to be falsy? If not, this seems more like a yaml validation.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    resources_data = data["resources"]
+    if "memory_gb" not in resources_data:
+        raise ValueError(f"Step '{step_name}': missing required 'memory_gb' in 'resources'.")
+    resources = ResourceConfig.from_dict(
+        resources_data, workflow_project=project, workflow_queue=queue
+    )
+    return step_name, resources
+
+
+def _require_args_block(data: dict[str, Any], step_name: str) -> dict[str, Any]:
```
rmudambi: How does this differ from `_require_arg()`?

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    """Return ``data['args']`` or raise ValueError when the block is missing."""
+    if "args" not in data:
+        raise ValueError(f"Step '{step_name}': missing required field 'args'.")
+    return cast(dict[str, Any], data["args"])
+
+
+def _require_arg(args: dict[str, Any], field: str, step_name: str) -> Any:
```
rmudambi: Since this actually returns the argument, I think `_get_arg()` is a better name.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    - A top-level ``command`` field always resolves to ``"command"``;
+      :func:`parse_command_step_from_yaml` enforces the rest of the
+      command-step schema (including any conflicting ``type``).
+    - Otherwise, an explicit ``type`` is used.
+    - Otherwise, ``"command"`` is the default, so the command-step parser
+      can raise a focused error about the missing ``command`` field.
+    """
+    if "command" in step_dict:
+        return "command"
+    step_type: str = step_dict.get("type", "command")
```
rmudambi: I don't think a default value makes sense here. If neither "command" nor "type" is in the step dict, we should error and say that you must provide a type or an explicit command with the command arg.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+        args["parameters"] = copy.deepcopy(parameters)
+    if cwd is not None:
+        args["cwd"] = str(cwd)
+
+    result["args"] = args
+    return result
+
+
+def _check_supported_args(
+    args: dict[str, Any], step_name: str, supported: set[str] | None
```
rmudambi: We never pass `None` to this function. I suspect we don't want that case.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+              memory_gb: 20
+              runtime: "02:00:00"
+              cores: 2
+    """
+    step_name, resources = _parse_common_step_fields(data, project=project, queue=queue)
+    _check_supported_args(data.get("args", {}), step_name, _COMMAND_SUPPORTED_ARGS)
+    if "command" not in data:
+        raise ValueError(f"Step '{step_name}': missing required field 'command'.")
+    explicit_type = data.get("type")
+    if explicit_type is not None and explicit_type != "command":
```
rmudambi:
```suggestion
    explicit_type = data.get("type", "command")
    if explicit_type != "command":
```

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+        raise ValueError(
+            f"Step '{step_name}': cannot specify both 'command' and "
+            f"'type: {explicit_type}'. When 'command' is set, 'type' "
+            "must be omitted or set to 'command'."
```
rmudambi: This sort of message is a bit confusing to me, since "command" could refer either to the step-type or an argument key. Maybe the "command" step-type should be changed to "shell" or "bash"?

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    output_directory: Path,
+    *,
+    project: str,
+    queue: str,
+) -> dict[str, Any]:
+    """Parse a raw pytest-step YAML dict into API kwargs.
+
+    Optional ``args`` keys: ``path``, ``k``, ``runslow``. At least one of
+    ``path`` or ``k`` must be provided (enforced by
+    :func:`~vivarium_cluster_tools.psimulate.workflow_config.validation.validate_pytest_step`).
```
rmudambi: You shouldn't have the comment saying where this is enforced imo

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    if parameters:
+        args["parameters"] = copy.deepcopy(parameters)
+    if cwd is not None:
+        args["cwd"] = str(cwd)
+
+    result["args"] = args
+    return result
+
+
+def _check_supported_args(
```
rmudambi: Nit: maybe better called `_check_for_unsupported_args()`?

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    data: dict[str, Any],
+    output_directory: Path,
+    *,
+    project: str,
+    queue: str,
+) -> dict[str, Any]:
+    """Parse a raw simulation-step YAML dict into API kwargs.
+
+    Required ``args`` keys: ``model_specification``, ``branch_configuration``.
+    Optional ``args`` keys: ``artifact_path``, ``backup_freq``,
```
rmudambi: I'm not convinced this comment is helpful, but if you are going to keep it, you definitely need to change ``args`` to ``data`` (or vice-versa), since it is very unclear what this refers to otherwise.

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
```diff
@@ -0,0 +1,186 @@
...
+            allow_none=False,
+        )
+
+
+def validate_command_step(
+    *,
+    name: str,
+    resources: ResourceConfig,
+    command: str,
+    output_directory: Path,
```
patricktnast: none of these validation functions use output directory or environment. why are they needed?

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+        kwargs["cwd"] = Path(args["cwd"]).resolve()
+    return kwargs
+
+
+def serialize_command_step_to_yaml(
+    *,
+    name: str,
+    resources: ResourceConfig,
+    command: str,
+    output_directory: Path,
```
patricktnast: similarly, none of these serialize functions use the output directory.

src/vivarium_cluster_tools/psimulate/workflow_config/task_builders.py
```diff
@@ -0,0 +1,282 @@
...
+
+if TYPE_CHECKING:
+    from jobmon.client.api import Tool
+    from jobmon.client.task import Task
+
+_NOTEBOOK_DEFAULT_KERNEL = "python3"
+"""Jupyter kernel used for notebook execution. Not user-configurable."""
+
+
+def build_command_step_tasks(
```
patricktnast: do you plan on using these for something besides to pass them into the interface? otherwise, i'm not sure why we need both modules.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+              positional_args:
+                - "foo"
+                - "bar"
+              keyword_args:
+                input_dir: /mnt/results/model_29
+                verbose: true
+                num_workers: 4
+    """
+    step_name, resources = _parse_common_step_fields(data, project=project, queue=queue)
+    args = copy.deepcopy(_require_args_block(data, step_name))
```
rmudambi: Why are we deepcopying? We shouldn't need the previous data again, so it doesn't matter if it gets mutated later (and I don't think it will anyway).

Same in other methods below

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+    if "environment" in data:
+        kwargs["environment"] = data["environment"]
+    if "parameters" in args:
+        kwargs["parameters"] = args["parameters"]
+    if "cwd" in args:
+        kwargs["cwd"] = Path(args["cwd"]).resolve()
+    return kwargs
+
+
+def serialize_command_step_to_yaml(
```
rmudambi: I think it's more intuitive if these methods take a `ParsedStep` object.

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
```diff
@@ -0,0 +1,186 @@
...
+    validate_common(name, resources)
+    if not command:
+        raise ValueError(f"Step '{name}': 'command' is required.")
+
+
+def validate_simulation_step(
+    *,
+    name: str,
+    resources: ResourceConfig,
+    output_directory: Path,
```
rmudambi: You're passing a lot of arguments through here which never get validated, such as `output_directory`, `environment`, `backup_freq`, etc. Either actually validate them, or don't pass them to this method.

Same for each of the validation methods below.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
```diff
@@ -0,0 +1,674 @@
...
+        project=resolved_project,
+        queue=resolved_queue,
+        output_directory=resolved_output_directory,
+        default_environment=workflow.get("default_environment"),
+        steps=steps,
+        max_attempts=max_attempts or workflow.get("max_attempts", DEFAULT_MAX_ATTEMPTS),
+    )
+
+
+def workflow_config_to_yaml_dict(config: WorkflowConfig) -> dict[str, Any]:
```
rmudambi: Nit: I think I prefer this as `workflow_config.to_dict()`. 

Also, maybe these serialization methods could be in a separate `serialization` module?

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
albrja: This is a current pain point of the API functions that I plan to update with https://jira.ihme.washington.edu/browse/MIC-7141. But we have workflow level configuration values we pass through to the API functions that the user also has to set here. I will remove them and add that back in with 7141.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: Good catch. This doesn't need to be here because it is workflow level configuration for the yaml.

src/vivarium_cluster_tools/psimulate/workflow_config/task_builders.py
albrja: Updated and renamed.

src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py
albrja: There is a functional difference environment = "some_name" and environment = None. We can't use positional arguments because if someone provided step name and an environment string (only 2 args), we do not know what the second argument is referencing because of None. 

However, we do not actually need the default_environment arg since it wasn't actually being used, we resolve it before this function so it is being removed.

src/vivarium_cluster_tools/psimulate/workflow_config/validation.py
albrja: ResourceConfig cannot be Falsy. And we are so this was just an extra conservative check and has been removed.

src/vivarium_cluster_tools/psimulate/workflow_config/task_builders.py
albrja: Removed module.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: I combined and renamed them since they didn't need to be seaprate. It is validating that the "args" block exists in the step dict in the yaml file and validating that required values in that block are there.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: We did pass None because of the command step not supporting any args block but since we went away from OOO we do not need to do that now.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: I think changing it to bash is a good idea.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: Removed.

src/vivarium_cluster_tools/psimulate/workflow_config/parsing.py
albrja: You are right it is not necessary.

patricktnast (approved):

rmudambi (approved):

PR #336, opened 2026-05-20
Albrja/feature/mic 7114/handle task logs

albrja:
### Add dual logging sink to all tasks
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7114

### Changes and notes
-add dual logging sink to all tasks
-manually parse task command args and do not rely on "--" (REMAINDER)

### Code reviewer
PR Review: Albrja/feature/mic 7114/handle task logs (PR #336)
Branches confirmed — feature albrja/feature/mic-7114/handle-task-logs → target albrja/refactor/mic-7115/wrap-api-functions (not main). Diff: 5 files, +317 / -25, 4 commits.

Summary
This PR adds a dual-stream logging mechanism so that failing typed-step tasks (pytest / python / notebook / command) surface their output in the SLURM stderr file and the Jobmon GUI. task_runner.py is reshaped into a two-mode CLI (simulation vs subprocess), where the new subprocess mode spawns a child process, mirrors its stdout (with stderr merged) to the parent's stdout in real time, buffers up to 10,000 lines in a deque, and on non-zero exit replays the buffer to stderr. A shared _configure_dual_sink() routes loguru INFO+ to stdout and WARNING+ to stderr.

BaseStepConfig gains a _WRAP_FOR_LOGGING: ClassVar[bool] = True flag and _wrap_for_logging() helper that prepends python -m …task_runner subprocess -- to each step's command inside _create_single_command_task. The pre-existing simulation entry point (workflow.py:get_task_list) is updated to pass the new simulation subcommand token, and one comment / two Usage:: block examples in the module docstring are rewritten.

Design
argparse.REMAINDER is non-deterministic and the test already hedges around it. [task_runner.py:113-117, 163-171](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L113-L171). REMAINDER's handling of a leading -- differs across CPython versions (bpo-17050). The test test_missing_double_dash_raises accepts either ValueError or SystemExit, which is a sign the author noticed the ambiguity. Replace with explicit slicing of argv, e.g. find the index of -- and slice — no REMAINDER needed. (flagged by Design, Maintainability, Tests, Documentation)

String-coupling of BaseStepConfig to the worker CLI shape. [config.py:331-333](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L331-L333) and [workflow.py:115](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py#L115). The literal "python -m vivarium_cluster_tools.psimulate.worker.task_runner …" appears in two production sites and 5 test sites. The docstring's reasoning ("avoid heavy work-horse imports during workflow-config parsing") is correct today — task_runner.py does pull in vivarium_work_horse at module scope — but it only justifies "don't import task_runner", not "don't share a constant". Suggested fix: add a thin worker/invocation.py (or use the existing worker/__init__.py, which is small) holding WORKER_MODULE = "..." plus subprocess_wrap(cmd) and simulation_prefix() helpers, lazy-import the work horses inside _run_simulation, then have both production sites and tests reference the constant. (flagged by Design, DRY, Maintainability)

stderr=subprocess.STDOUT + replay-only-on-failure silently drops child warnings on success. [task_runner.py:175-181, 189-192](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L175-L192). Because child stderr is merged into stdout and the buffer is only replayed on non-zero exit, any loguru.WARNING or warnings.warn emitted by the child during a successful run never reaches the parent's stderr → never reaches the SLURM stderr file or Jobmon GUI stderr pane. Given the PR's stated motivation is making warnings/errors visible to the GUI, this is the opposite of the intended behavior on the success path. Consider draining child stderr separately (two threads or selectors.DefaultSelector) and forwarding each stream live, with the replay buffer reserved for failure amplification.

No child-process lifecycle management on parent termination. [task_runner.py:175-187](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L175-L187). When SLURM hits walltime or scancel fires, the parent gets SIGTERM. There is no try/finally, no proc.terminate(), no signal forwarding — so the child can be orphaned (until SLURM reaps the whole cgroup), and during the grace window the child can block on a full pipe (its traceback never makes it to disk). Wrap the read loop in try/finally, add signal.signal(SIGTERM, …) handlers that forward to proc.send_signal(), and call proc.terminate()/wait(timeout=…)/kill() in the finally. This is the failure mode where surfacing output matters most, so leaving it unhandled undercuts the PR's value.

Buffer is tail-only. [task_runner.py:52-54, 174](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L52-L174). deque(maxlen=10_000) keeps the most recent lines, but for chained Python tracebacks the root cause is at the head and the summary is at the tail. Consider a small head-list + tail-deque pattern with an "X lines elided" separator. Optional but cheap.

_WRAP_FOR_LOGGING opt-out is dead code. [config.py:191-196](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L191-L196) advertises an opt-out for "simulation steps that already have their own dual-sink logging setup". But SimulationStepConfig (config.py:473+) overrides get_tasks entirely and never calls _create_single_command_task, so the flag is unused by the only intended consumer. Two viable resolutions: (a) YAGNI — drop _WRAP_FOR_LOGGING and _wrap_for_logging's opt-out branch until a real subclass needs it; or (b) actually set _WRAP_FOR_LOGGING = False on SimulationStepConfig so the documented contract has a real user, then add a regression test. (flagged by Design, Maintainability, Tests, Documentation)

assert proc.stdout is not None is type-narrow, not runtime safety. [task_runner.py:182](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L182). Disappears under python -O. Replace with typing.cast(IO[str], proc.stdout) or a tiny comment explaining the narrow is for mypy and is guaranteed by stdout=PIPE above.

Maintainability
_configure_dual_sink docstring overstates "default" handler removal. [task_runner.py:57-67](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L57-L67). The docstring says "Removes loguru's default stderr handler first…" but the code calls logger.remove() with no argument, which removes every registered sink. Today the difference is invisible (no other code path adds a sink before this runs), but a future caller (test fixture, parent script) that registers a sink first will get silently obliterated. Either narrow to logger.remove(0) (matches the docstring) or update the docstring to say "all currently-registered sinks". (flagged by Maintainability, Documentation)

Inline comment at [task_runner.py:112](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L112) is WHAT, not WHY, and slightly misleading. # Everything after --is collected by argparse intoinner_argv. describes the code rather than the reason for using REMAINDER, and REMAINDER actually includes the -- token (which is why _run_subprocess strips it). Replace with the WHY — or, if Design #1 is adopted, the comment becomes moot.

bufsize=1 + text=True are load-bearing for real-time mirroring and undocumented. [task_runner.py:175-181](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L175-L181). Without these, for line in proc.stdout waits for buffer-fill before each iteration. A future refactor that "tidies up" by dropping bufsize=1 would silently lose the real-time-mirror property without breaking any test (the existing tests use tiny outputs that fit in one flush). One short comment: # bufsize=1 + text=True → line-buffered iteration so we mirror output as the child produces it.

logger.info(f"Running subprocess: {' '.join(inner_argv)}") is shell-ambiguous. [task_runner.py:173](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L173). For an argv like [PY, "-c", "print('hello world')"], the logged line becomes python -c print('hello world') — not pasteable to a terminal. Use shlex.join(inner_argv).

BUFFER_MAXLEN test-monkeypatching contract is implicit. [task_runner.py:52-54](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L52-L54). The constant is read inside _run_subprocess via a module-globals lookup at call time, which is what makes monkeypatch.setattr(task_runner, "BUFFER_MAXLEN", 5) work. A future "from .task_runner import BUFFER_MAXLEN" elsewhere or even a local rebind silently breaks the test. Either inject the cap as a kwarg (def _run_subprocess(..., *, max_lines: int = BUFFER_MAXLEN)) or accept the coupling but document it explicitly in the constant's docstring.

DRY
Module-path string duplicated across 2 production sites + 5+ test sites. Same finding as Design #2 — addressing it removes the DRY violation. Specifically: [workflow.py:115](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py#L115), [config.py:332](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L332), and test sites at [test_config.py:382, 438, 461, 474, 949](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py#L438).

Repeated substring assertion "vivarium_cluster_tools.psimulate.worker.task_runner subprocess --" appears at [test_config.py:438, 461, 474](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py#L438). Extract a _WRAPPER_PREFIX constant (cascades naturally from DRY #1).

["subprocess", "--", PY, "-c", <snippet>] argv literal repeats in 7 of 8 TestSubprocessMode methods. [test_task_runner.py:215, 223, 234, 240, 243, 249-257, 268-276](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/worker/test_task_runner.py#L215). The existing _build_argv helper for simulation mode sets a precedent — add a sibling def _subprocess_argv(code: str) -> list[str]: return ["subprocess", "--", PY, "-c", code]. The negative test (test_missing_double_dash_raises) is intentionally a different shape and should stay literal.

test_pytest_step_command_is_wrapped_with_runner and test_command_step_command_is_wrapped_with_runner are near-duplicates (18 lines each, differ only in config kwargs and inner-command substring). [test_config.py:418-462](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py#L418-L462). Parametrize over (make_config, inner_substring).

Tests
_configure_dual_sink has no direct test. Its level-routing contract — INFO+ → stdout, WARNING+ → stderr, no INFO duplicated on stderr — is observable but unasserted. Every TestSubprocessMode test uses sys.stdout.write/sys.stderr.writelines paths that bypass loguru, so a regression like flipping level="WARNING" to level="ERROR" or forgetting logger.remove() would slip through. Add a small direct test that emits records at multiple levels and asserts which streams they land on. (must fix)

The empty-argv-after--- branch is untested. [task_runner.py:170-171](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L170-L171) raises ValueError("subprocess mode requires argv after the '--' separator.") but no test exercises it (test_missing_double_dash_raises exercises the missing--- branch only). Add main(["subprocess", "--"]) → pytest.raises(ValueError, match="after the '--' separator"). (must fix)

test_failure_replays_buffered_output_to_stderr does not assert the new ERROR log line. [test_task_runner.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/worker/test_task_runner.py). The test only checks "boom" in err, missing the logger.error("Subprocess exited with code 2; replaying output to stderr.") emission that is the only observable signal that _configure_dual_sink's WARNING+ sink is wired correctly. Add assert "Subprocess exited with code 2" in err. (must fix)

The simulation-mode worker_command template is not regression-protected. The PR adds the simulation token to [workflow.py:115](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py#L115) — without it, every simulation task fails at runtime because parse_args requires mode. But tests/psimulate/jobmon_config/test_workflow.py does not assert the command_template. Add a unit assertion that "task_runner simulation " appears in the command_template kwarg passed to tool.get_task_template. (must fix)

test_missing_double_dash_raises accepts both ValueError and SystemExit. [test_task_runner.py:282-286](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/worker/test_task_runner.py#L282-L286). The implementation deterministically raises ValueError for this input, so pin the assertion to ValueError with a match=... clause. This dovetails with Design #1.

Wrap-with-runner substring assertions don't enforce ordering. [test_config.py:437-439, 460-462](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/workflow_config/test_config.py#L437-L462). A regression that flipped the order (echo hello world python -m … subprocess --) would still pass. Strengthen to a single contains check: assert "subprocess -- echo hello world" in cmd.

test_buffer_cap_does_not_crash_on_large_output assertions are weaker than they could be. [test_task_runner.py:260-280](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/psimulate/worker/test_task_runner.py#L260-L280). Sharpen to assert f"line-{n}\n" in err for n in 95..99 and assert f"line-{n}\n" not in err for n in 0..94, so the exact cap behavior — including that the tail and not the head is retained — is documented by assertion. Also assert code == 1 upfront.

No test exercises the child-stderr-only path. Given stderr=subprocess.STDOUT, child stderr is merged into stdout and replayed on failure. The most common real failure mode (Python traceback to stderr) isn't tested. Add one.

No coverage for the implicit SimulationStepConfig opt-out. Connected to Design #6 — either set the flag and assert via a unit test that simulation-step commands are NOT wrapped, or add a comment-only test noting the current bypass path.

Documentation
task_runner.py module docstring uses "workflow log file" — not a name used anywhere else in the repo. [task_runner.py:20](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L20). The existing docs/source/logging.rst uses cluster_logs (SLURM stdout/stderr) and worker_logs. Suggest changing to "(the per-task SLURM stdout file, i.e. under cluster_logs/)" so cross-doc search lands users in the right place.

_run_subprocess docstring omits the buffer cap. [task_runner.py:154-162](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L154-L162). The user-visible contract — "only the last 10,000 lines are replayed on failure" — should be stated explicitly so an operator hunting a failure knows the GUI shows the tail, not the whole transcript.

_WRAP_FOR_LOGGING class-attribute docstring's example is wrong. [config.py:191-196](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L191-L196). It cites "simulation steps that already have their own dual-sink logging setup" as the opt-out example, but SimulationStepConfig never sets the flag — it bypasses the wrapping path entirely via overriding get_tasks. Either rewrite the example or, better, make it real (see Design #6).

_wrap_for_logging docstring's import-avoidance reasoning would survive a refactor better if specific. [config.py:316-327](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/config.py#L316-L327). Strengthen "heavy work-horse imports" to "would transitively pull in vivarium.framework.engine and dill via vivarium_work_horse". Optional.

No CHANGELOG.rst entry for the dual-sink behavior. The open 3.2.0 - TBD section has one bullet (the workflow command). Externally-relevant changes from this PR — task stderr now contains failing-command output; task_runner CLI now requires a simulation/subprocess subcommand — should be queued for the 3.2.0 changelog before the cumulative branch lands on main.

PR body's "testing" section failed to load (per WebFetch). Since commit messages are terse (Add tests and code stubs, Implementation for dual logging sink, Move helper method, two Merges), the PR body is the only place a reviewer can read the motivation, the design choice to multiplex via subprocess -- rather than per-step wrapping, and the buffer-cap trade-off. Worth ensuring the body actually renders before requesting review.

Module docstring Usage:: block uses \\ line continuations. [task_runner.py:24-33](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L24-L33). Renders as single \ after Sphinx — fine, just flagging that it looks like a typo in raw form.

Functionality
Behavioral change: the task_runner CLI now requires a subcommand. Any pre-PR Jobmon task command persisted on disk (e.g., a paused/resumed workflow whose tasks were created before this PR) will fail when re-executed, because parse_args now requires mode. Internal callers are updated (the only one is workflow.py:115), but if a workflow is mid-flight at deploy time and is resumed afterward, the tasks already in Jobmon's database invoke the old form. This is the same class of risk as a CLI-flag-rename. Worth confirming the deployment story (e.g., document a "no in-flight workflows" precondition for upgrade, or add a back-compat default subcommand).

main() signature changed from -> None to -> int and __main__ now sys.exit(main()). The only in-process importer is the test module, which doesn't check the return value, so this is safe. Worth noting in the PR description.

Order of stderr emissions on failure is: loguru ERROR line → buffered child output. [task_runner.py:189-192](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L189-L192). Reasonable, but means anyone scrolling to the top of the SLURM stderr file sees the runner's wrapper message before the actual failure transcript. Operator-friendly; flagging only for awareness.

Silent-stderr edge case when buffer is empty AND exit_code != 0 — e.g., child exits with code 2 having produced no output. Stderr will contain only the runner's ERROR line and no transcript to debug. Acceptable but worth a one-line note in the docstring or a different ERROR message ("Subprocess exited with code N with no output captured.").

text=True will raise UnicodeDecodeError on a child that emits invalid UTF-8 (e.g., a binary blob accidentally written to stdout, or a non-UTF-8 locale). The exception propagates out of the for line in proc.stdout loop, the child is leaked (no cleanup; see Design #4), and Jobmon sees a Python traceback instead of the wrapped task's exit code. Either pass errors="replace" to Popen (Python 3.7+) or wrap the read loop in try/except. Cheap, robust fix.

Minor Nits
Negate _WRAP_FOR_LOGGING polarity to _BYPASS_LOGGING_WRAPPER: ClassVar[bool] = False — reads correctly inline (if self._BYPASS_LOGGING_WRAPPER: return command) and doesn't require the docstring to explain "True means wrap". Stylistic.

_configure_dual_sink() is called after parse_args in main(). [task_runner.py:196-198](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/task_runner.py#L196-L198). Argparse errors therefore bypass loguru. Argparse writes to stderr directly so nothing is lost, but the docstring says "at the top of each worker entry point" — it's actually second. Either swap the order or relax the docstring wording.

Test name test_buffer_cap_does_not_crash_on_large_output understates the contract. Rename to test_buffer_cap_preserves_tail_drops_head so a future maintainer who swaps deque for a different capped buffer is alerted that head/tail polarity matters.

test_argv_after_double_dash_is_executed does not assert the exit code. Add assert main([...]) == 0 so a future regression that returns -1 from a successful subprocess is caught.

Overall
The PR has a clear goal and the test suite for the new code path is reasonably thick. The two changes most worth doing before merge are (1) replacing argparse.REMAINDER with explicit slicing — the test's pytest.raises((ValueError, SystemExit)) is a smell that the current parsing is unreliable across CPython versions; and (2) adding child-process lifecycle handling (try/finally + SIGTERM forwarding), because SLURM walltime is the primary failure mode this PR is meant to make debuggable, and right now that path orphans the child mid-write. Adding a single source of truth for the worker module path (Design #2 / DRY #1) is the next-highest-value follow-up; without it, the parent refactor branch's whole point — moving things — has two unrelated string literals to grep on each move.

Two design questions worth a conversation rather than a unilateral fix: whether warnings during successful child runs should reach the Jobmon GUI (currently they don't), and whether the _WRAP_FOR_LOGGING opt-out should be deleted (YAGNI) or realized by setting False on SimulationStepConfig. Both are coherent design choices — the docstring just needs to match whichever the author intended.

Test gaps to close before merge: a direct test for _configure_dual_sink, the empty-argv-after--- branch, the new logger.error failure-path emission, and a command_template regression for the simulation worker_command in test_workflow.py.


src/vivarium_cluster_tools/psimulate/worker/task_runner.py
```diff
@@ -3,23 +3,41 @@
 Jobmon Task Runner
 ========================
 
-CLI entry point for individual Jobmon tasks. Each task loads its metadata
-JSON file, runs the appropriate work horse, and writes results directly
-to the results directory (one parquet per metric per task).
+Single CLI entry point for Jobmon worker tasks. Two execution modes,
+dispatched by the first positional argument:
+
+* ``simulation`` — load a task's metadata JSON, run the appropriate work
```
rmudambi: What would happen if we subprocessed the simulation steps?

src/vivarium_cluster_tools/psimulate/worker/task_runner.py
```diff
@@ -95,7 +139,77 @@ def main(argv: list[str] | None = None) -> None:
...
+            except subprocess.TimeoutExpired:
+                proc.kill()
+                proc.wait()
+        if proc.returncode != 0:
+            logger.error(
+                f"Subprocess exited with code {proc.returncode}; "
+                "replaying output to stderr."
+            )
+            sys.stderr.writelines(buffered)
+            sys.stderr.flush()
```
rmudambi: Do we really have to do all this manually? It seems like the sort of thing we should be able to rely on a logging library to do for us?

src/vivarium_cluster_tools/psimulate/worker/task_runner.py
albrja: I didn't want to change the behavior or run/restart but it won't so I will do it to be consistent.

src/vivarium_cluster_tools/psimulate/worker/task_runner.py
albrja: This isn't really a logging library concern but its necessary to make the logging update work and address one of the issues the code reviewer raised. I have updated the doc string to hopefully clear this up.

albrja: It does look like it but since this is a subprocess, the child's stdout is a raw byte stream from another OS process, not records. We would have to read the pipe, then wrap each line in a LogRecord which is the reason this wasn't showing up in the Jobmon GUI

patricktnast:
It would be nice to see an example of the logging stream /  files that this change produces relative to previous.

albrja:
@patricktnast Before it only showed the command section in the Jobmon GUI
image

src/vivarium_cluster_tools/psimulate/workflow_config/utilities.py
```diff
@@ -121,7 +121,14 @@ def get_single_command_task(
...
     command: str,
 ) -> list[Task]:
-    """Return a one-element ``list[Task]`` for a step that runs a single command in a conda env."""
+    """Return a one-element ``list[Task]`` for a step that runs a single command in a conda env.
+
+    The child inherits SLURM's stdout/stderr file descriptors via the
+    ``stdout`` / ``stderr`` keys on the compute resources dict
+    (configured by :meth:`NativeSpecification.to_jobmon_spec`), so its
+    output flows directly to the per-step log files that the Jobmon GUI
+    surfaces.
```
rmudambi: No need for this comment.

src/vivarium_cluster_tools/psimulate/worker/task_runner.py
```diff
@@ -95,7 +102,8 @@ def main(argv: list[str] | None = None) -> None:
         results_dict=results_dict,
     )
     logger.info(f"Task {task_id} results written successfully.")
+    return 0
 
 
 if __name__ == "__main__":
-    main()
+    sys.exit(main())
```
rmudambi, line 109: Is this basically the change that matters? Or is it the removal of the default loguru handler?

tests/psimulate/jobmon_config/test_workflow.py
rmudambi: What purpose does this method serve? Can't we just call `build_workflow()` directly when we need it?

rmudambi (approved):

patricktnast (approved):

src/vivarium_cluster_tools/psimulate/worker/task_runner.py
albrja: It is a combination of both. This doesn't change the simulation tasks but we did need to remove the loguru handler for other step types.

tests/psimulate/jobmon_config/test_workflow.py
albrja: Removed

PR #340, opened 2026-05-26
Albrja/feature/mic 6938/no time increase

albrja:
### Reduce runtime resource on restart with backups
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-6938

### Changes and notes
-cut runtime resource in half for simulations if doing a restart and backups are present


albrja:
Code Reviewer
PR Review: Cut runtime in half on simulation retries when backup enabled (PR #340)
Summary
PR 340 changes Jobmon's per-task retry-scaling policy for simulation tasks. When a JobParameters has backup_freq set (i.e., backups are enabled), the worker resumes from the most-recent backup pickle, so a retry needs less wall time than the original allocation. The PR makes simulation-task retries request resource_scales = {"memory": 0.5, "runtime": -0.5}, which (under Jobmon's ceil(orig * (1 + scale)) formula) keeps memory growing 1.5× on each retry (matching Jobmon's hidden default) but shrinks runtime to 0.5×. When backup_freq is None, no resource_scales is forwarded and Jobmon's built-in default of +50% memory / +50% runtime continues to apply.

The change spans four files: client.py (forward resource_scales through the facade), workflow.py (new private _simulation_resource_scales helper called inside get_task_list), conftest.py (fixture now uses a real BackupConfiguration with backup_freq=None instead of the empty dict that previously would have raised KeyError), and test_workflow.py (new TestGetTaskListBackupAwareRetryScaling class with two cases). ~119 net lines added across the two substantive commits.

Design
Helper coupled to internal shape of JobParameters, when both callers have the flag in scope — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:163-184. _simulation_resource_scales reaches into job_parameters_list[0].backup_configuration["backup_freq"]. Both call sites already have backup_freq as a local: [interface.py::get_simulation_step_tasks](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/workflow_config/interface.py#L111) takes it as a parameter, and runner.py::build_workflow has it in scope. A backups_enabled: bool kwarg on get_task_list would (a) make the contract explicit at the function boundary, (b) eliminate the IndexError on empty list (see Functionality #1), (c) let the test class drop the contrived BackupConfiguration rigging and just toggle a bool, and (d) decouple the helper from the specific shape of BackupConfiguration. [Flagged by Design.]

Conditional-kwarg pattern in create_tasks is defensive but not behaviorally required — src/vivarium_cluster_tools/psimulate/jobmon_config/client.py:123-132. Verified against Jobmon source: [array.py:83-84](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/index.html?id=1d16f169-1f45-4dab-b839-558c26a6de6a&parentId=1&origin=0cfe4aad-c7bb-4468-aebe-3336c126e99b&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&remoteAuthority=wsl%2Bubuntu&session=cfcef968-c5e4-4e0b-9747-0c536325c92f) does resource_scales if resource_scales is not None else {} — passing None is observably identical to omitting the kwarg. A future maintainer reading client.create_tasks cannot tell that from the local code, and test_backups_disabled_omits_resource_scales couples to the conditional-insert implementation rather than to Jobmon-level behavior. Either keep the pattern (it matches make_task_template / make_workflow precedent) and document the rationale, or simplify to always-forward and update the test to assert kwargs.get("resource_scales") is None. [Flagged by Design, Maintainability, Tests.]

Good design choice: retry policy is intrinsic, not user-tunable — Keeping the policy derived from backup_freq (rather than exposing a CLI flag --no-retry-scale-down) is correct: the property is a worker-implementation invariant. The worker itself checks backup_freq is not None to decide whether to resume from backup ([vivarium_work_horse.py:111](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py#L111)). Exposing it would invite users to set values that diverge from worker behavior. [Confirmed by Design.]

Good design choice: only simulation tasks are affected — Other workflow step types (bash, pytest, python, notebook) still get Jobmon's default 1.5× runtime escalation. This is correct — only the simulation worker has backup-aware resume semantics. [Confirmed by Design.]

Maintainability
-0.5 rationale lives in the PR description, not the code — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:181-184. The inline comment explains the formula and the sign convention, but not why -0.5 specifically — a future reader sees an arbitrary constant. A one-line addition tying the choice to "typical backup_freq is ~30 min, so half the runtime is comfortably above the largest 'since last backup' interval" would document the conditions under which the constant becomes wrong (e.g., very short total runtimes where the backup interval approaches half the allocation). [Flagged by Maintainability, Design.]

Inline comment is correct but "matching Jobmon's default behavior" can be misread — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:181-183. The parenthetical applies to memory only; runtime is diverging from Jobmon's default. Suggested rewrite:


# Jobmon's scale formula is ceil(orig * (1 + scale)). +0.5 on memory
# keeps Jobmon's default 1.5x retry growth; -0.5 on runtime shrinks
# the retry's request to half because the worker resumes from backup.
[Flagged by Maintainability, Documentation.]

Helper name elides its conditional/policy nature — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:163. _simulation_resource_scales reads as a getter; the function is actually answering "given backup state, what retry policy applies?" Renaming to _resolve_retry_scales or _retry_scales_for_backup_policy would signal the conditional. [Flagged by Maintainability.]

Real-looking paths in fixture default could mislead test authors — tests/psimulate/conftest.py:27-31. The default backup_dir="/tmp/backups" and backup_metadata_path="/tmp/backup_metadata.csv" look like writable filesystem paths but are never touched in any test that uses the default fixture. A future author may try to write through them. Consider sentinel-looking strings ("<unused-backup-dir>") or Path("/nonexistent/..."), plus a one-line comment that these paths are unused in the default-backups-disabled case. [Flagged by Maintainability.]

Docstring references worker module by Sphinx role — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:170. The :func: cross-ref to vivarium_work_horse.get_backup resolves today, but creates an implicit dependency that won't break a test if get_backup is renamed or moved. Acceptable trade-off for traceability; just be aware. [Flagged by Maintainability.]

DRY
_call_get_task_list static helper is duplicated between two test classes — tests/psimulate/jobmon_config/test_workflow.py:224-242 and 305-320. The two static helpers call get_task_list with the same forwarded args; the only difference is the wrapping-test version threads through wrap_command. Either lift one helper to module scope or have the new class re-use the existing one with wrap_command=False. This is a real maintenance burden — get_task_list's signature changed in this very PR, and both helpers had to be updated in lockstep. [Flagged by DRY.]

BackupConfiguration(...) literal appears in both conftest and the new test — tests/psimulate/conftest.py:27-31 and tests/psimulate/jobmon_config/test_workflow.py:335-339. Both share backup_dir / backup_metadata_path and differ only in backup_freq. A make_backup_configuration(**overrides) helper next to make_job_parameters would (a) eliminate the duplication, (b) be reusable by the other test files that already construct BackupConfiguration literals. Borderline today, but the pattern will multiply. [Flagged by DRY, Tests.]

Conditional-kwarg pattern matches existing precedent in client.py — client.py:70-80, 123-132, 149-157, 192-195 all use the same "build dict, maybe add key, splat" pattern for optional Jobmon SDK kwargs. The new code follows the established convention; extracting a _drop_none helper would harm clarity. Not a finding — flagging only as confirmation. [Confirmed by DRY.]

Tests
No direct unit tests for client.create_tasks facade — tests/psimulate/jobmon_config/ lacks a test_client.py. The conditional resource_scales forwarding is only covered transitively through get_task_list. A regression in client.create_tasks (e.g., dropping the kwarg, renaming it, switching to always-forward) cannot be distinguished from a regression in _simulation_resource_scales. Suggested two-test addition shown in the test reviewer's report. [Flagged by Tests.]

No unit tests for _simulation_resource_scales directly — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:163-184. Uncovered behaviors: empty list (IndexError), backup_freq=0 (treated as "enabled" by is not None), negative backup_freq, mixed backup_configuration across the list. A dedicated TestSimulationResourceScales class would pin each behavior. [Flagged by Tests.]

test_backups_disabled_omits_resource_scales is asymmetric with its sibling — tests/psimulate/jobmon_config/test_workflow.py:351-369. The enabled-case asserts exact dict equality; the disabled-case only asserts absence. This couples the disabled test to the conditional-insert implementation; if create_tasks is refactored to always-forward resource_scales=None, the test breaks despite identical Jobmon behavior. Consider assert kwargs.get("resource_scales") is None. [Flagged by Tests, Design.]

test_backups_enabled_halves_runtime_on_retry is a tautological mirror of the helper literal — tests/psimulate/jobmon_config/test_workflow.py:341-348. If someone changes the helper to {"memory": 0.6, "runtime": -0.4}, both update together with no signal that intent drifted. Tie the values to semantics — assert runtime < 0 ("retry should not increase runtime when a backup exists") and document why memory=0.5 matches Jobmon's default. [Flagged by Maintainability, Tests.]

build_workflow path (the production psimulate run/restart entry point) is not asserted for the new behavior — tests/psimulate/jobmon_config/test_workflow.py::TestBuildWorkflow. test_tasks_added_to_workflow (lines 193-218) already inspects task_template.create_tasks.call_args.kwargs; extending it with one line (assert "resource_scales" not in kwargs) — or adding a parallel test that goes through build_workflow with backups enabled — would pin the CLI path. Currently, if build_workflow ever stops calling get_task_list (or wraps it differently), the feature breaks on the actual CLI path while only the lower-level test still passes. [Flagged by Tests.]

Conftest fixture change from {} to BackupConfiguration(...) silently changes the default test policy — tests/psimulate/conftest.py:27-31. The change is necessary (the old {} would raise KeyError against the new helper), but tests that previously failed loudly on missing config now silently get a "backups disabled" path. Consider a docstring note flagging that the default represents "backups disabled", or two named factories (make_job_parameters_with_backups / _without_backups) to make intent explicit at call sites. [Flagged by Tests.]

Deep mock-chain assertions are brittle to facade refactors — tests/psimulate/jobmon_config/test_workflow.py:346-348, 366-368. Both new tests reach 5 levels deep through mock_tool_cls.return_value.get_task_template.return_value.create_tasks.call_args.kwargs. If anyone refactors the facade to patch client.make_task_template separately, both tests break for reasons unrelated to backup-aware scaling. A direct mocker.patch("...workflow.client.create_tasks") would make the assertion shallower and the test more focused. [Flagged by Tests.]

Documentation
No CHANGELOG entry for a user-visible behavior change — CHANGELOG.rst. PR 340 does not add an entry. When users set --backup-freq, retried simulation tasks now request ceil(orig_runtime * 0.5) instead of ceil(orig_runtime * 1.5). This can flip whether a retry lands in a different SLURM queue or fits within remaining runner-node lifetime. Suggested wording: "When --backup-freq is set, simulation-task retries now request half the original runtime (worker resumes from the most-recent backup pickle); without backups, Jobmon's default 1.5x runtime escalation continues to apply." [Flagged by Documentation.]

create_tasks docstring does not mention the new resource_scales parameter — src/vivarium_cluster_tools/psimulate/jobmon_config/client.py:113-122. The parameter was added but the docstring's Parameters section was not updated. Also has a dangling blank line before the closing """ (inconsistent with the docstring style of sibling functions like create_task). Add:


resource_scales
    Optional per-resource retry scaling. When omitted, Jobmon falls
    back to its built-in default of ``{"memory": 0.5, "runtime": 0.5}``
    (i.e., +50% on each retry).
[Flagged by Documentation, Maintainability.]

distributed_runner.rst is not updated, but did not previously document retry scaling — docs/source/distributed_runner.rst:46-47, 55-59 describes --backup-freq and --max-runtime independently. Adding "When set, failed simulation tasks that are retried request half the original runtime" under --backup-freq would be the natural place if user-facing documentation is desired. Not required since the prior behavior was also undocumented. [Flagged by Documentation.]

"Representative-of-batch" claim is accurate today but not enforced — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:175-177. The docstring asserts callers always pass a homogeneous list; both current callers do, but a future caller mixing per-job backup configurations would silently get a wrong policy for half the batch. Either add assert len({jp.backup_configuration["backup_freq"] for jp in job_parameters_list}) == 1 (loud failure on violation) or soften the docstring to "the helper assumes a homogeneous list and reads only the first entry". [Flagged by Maintainability, Design, Documentation, Tests.]

Functionality
IndexError on empty job_parameters_list — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:179. Prior to this PR, an empty list yielded an empty task list (get_task_list would still execute the metadata loop trivially and create_tasks would no-op). Now the helper crashes first with a cryptic IndexError from a function whose name gives no hint that emptiness is forbidden. Either guard with if not job_parameters_list: return None, assert non-emptiness at the top of get_task_list, or — better — adopt the caller-passed-flag refactor from Design #1 which eliminates the concern entirely. [Flagged by multiple reviewers; my analysis confirms the previous behavior was "no-op", so this is a silent regression in error-mode robustness.]

Scaling compounds across retries — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:184. Jobmon's adjust_resources ([task_resources.py:175](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/index.html?id=1d16f169-1f45-4dab-b839-558c26a6de6a&parentId=1&origin=0cfe4aad-c7bb-4468-aebe-3336c126e99b&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&remoteAuthority=wsl%2Bubuntu&session=cfcef968-c5e4-4e0b-9747-0c536325c92f)) scales from existing_resources (the most recent allocation), not the original. With default max_attempts=3, a sim that times out twice would get:

Attempt 1: 24h
Attempt 2: ceil(24h × 0.5) = 12h
Attempt 3: ceil(12h × 0.5) = 6h
For a sim that fails before any backup is written (e.g., setup error, transient infra failure in the first few minutes), the worker correctly handles "no backup" (restarts from scratch and logs a warning at [vivarium_work_horse.py:121-122](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/worker/vivarium_work_horse.py#L121-L122)), but on retry it has only half the time — and on a second retry, only a quarter. This can produce cascading failures for short-lived setup errors. Worth either:

documenting this as a known trade-off (the user previously accepted "ship -0.5 as-is, tune later" in the design discussion),
or using a stop-iter / iterable-based scale (e.g., iter([-0.5])) so subsequent retries don't compound.
backup_freq=0 would be treated as "backups enabled" — src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py:179. The check is is None, so 0 (and 0.0) flow through as enabled. Real-world impossibility (no production code paths construct this), but worth flagging since the helper doesn't validate. The CLI layer doesn't appear to bound backup_freq to positive values.

None vs. omit are equivalent in Jobmon, so the defensive kwarg pattern is style-only — Confirmed by reading [array.py:83-84](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/index.html?id=1d16f169-1f45-4dab-b839-558c26a6de6a&parentId=1&origin=0cfe4aad-c7bb-4468-aebe-3336c126e99b&swVersion=5&extensionId=Anthropic.claude-code&platform=electron&vscode-resource-base-authority=vscode-resource.vscode-cdn.net&parentOrigin=vscode-file%3A%2F%2Fvscode-app&remoteAuthority=wsl%2Bubuntu&session=cfcef968-c5e4-4e0b-9747-0c536325c92f). See Design #2.

Minor Nits
Trailing blank line in create_tasks docstring — src/vivarium_cluster_tools/psimulate/jobmon_config/client.py:121. Inconsistent with create_task (no trailing blank). [Flagged by Documentation.]

native_spec fixture lacks a docstring while the other fixtures in test_workflow.py have them — tests/psimulate/jobmon_config/test_workflow.py:56-66. Pre-existing, not from this PR. [Flagged by Documentation.]

Overall
The core mechanism is sound and the implementation is small (~119 net lines). The biggest design concern is the introspection-based "backups enabled" lookup inside _simulation_resource_scales: it couples get_task_list to the shape of JobParameters.backup_configuration, silently changes the error mode on empty input from "no-op" to "IndexError", and forces tests to construct contrived BackupConfiguration rigging when a backups_enabled: bool kwarg would be cleaner at every level. Both callers have the flag in scope.

The next-most-impactful gaps are the missing CHANGELOG entry for a user-visible behavior change, the missing direct tests for client.create_tasks (and for _simulation_resource_scales itself), and the compounding-scale behavior across multiple retries — a sim that fails twice with backups enabled will get 0.25× of its original runtime on the third attempt, which deserves at minimum an in-code note.

The test class has a clear structure and good docstrings, but the duplicated _call_get_task_list helper and the asymmetric assertion style (exact-equality enabled, absence-only disabled) are worth tightening. The create_tasks docstring should mention the new resource_scales parameter.

Recommended pre-merge actions, in order: (1) CHANGELOG entry, (2) document resource_scales in create_tasks docstring, (3) clarify the "matches Jobmon's default" inline comment, (4) decide on the caller-flag refactor (Design #1) vs. adding an empty-list guard. The remaining items are good follow-ups.

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -148,11 +148,28 @@ def get_task_list(
...
+
+    Jobmon's scale formula is ``ceil(orig * (1 + scale))``. Supplying any
+    ``resource_scales`` dict suppresses Jobmon's full default, so memory
+    is restated at ``0.5`` to preserve its default +50% retry growth;
+    runtime is set to ``-0.5`` to shrink the retry's request to half
+    because the worker resumes from backup.
+    """
+    if job_parameters_list[0].backup_configuration["backup_freq"] is None:
+        return None
+    return {"memory": 0.5, "runtime": -0.5}
```
albrja, line 170: This is only decreasing the runtime by 50% on each retry. Memory is STILL increasing by 50% on each retry which is the default for Jobmon but Jobmon treats this dict as an all or nothing dict that has to be provided.

CHANGELOG.rst
```diff
@@ -2,6 +2,10 @@
 
 - Implement serial workflows via "psimulate workflow" command
 - Add facade to interact with Jobmon SDK
+<<<<<<< HEAD
+- Reduce simulation task runtime resource during restart with backups
+=======
+>>>>>>> epic/phase-3-automated-validation
```
rmudambi: Need to fix this failed merge resolution

rmudambi (approved):

patricktnast:
The unit tests check the resources dictionary but we don't (possibly can't?) check whether jobmon actually scales the job correctly. did you test this somehow? The format you use here is somewhat different from the documentation (which could be outdated) https://jobmon.readthedocs.io/en/latest/user_guide/compute_resources.html#custom-scaling

albrja:
@patricktnast The code lets you do float, callable, or iterator - https://github.com/ihmeuw-scicomp/jobmon/blob/main/jobmon_client/src/jobmon/client/task_resources.py#L181

src/vivarium_cluster_tools/psimulate/jobmon_config/workflow.py
```diff
@@ -148,11 +148,28 @@ def get_task_list(
...
     return tasks
 
 
+def _simulation_resource_scales(
+    job_parameters_list: list[JobParameters],
+) -> dict[str, float] | None:
+    """Per-task retry scaling for simulation tasks.
+
+    Jobmon's scale formula is ``ceil(orig * (1 + scale))``. Supplying any
+    ``resource_scales`` dict suppresses Jobmon's full default, so memory
```
hussain-jafari, line 163: Nit: I think something like "memory is defined with the default value explicitly" would be more clear.

hussain-jafari (approved):

PR #341, opened 2026-05-27
Albrja/feature/mic 7066/new cli group

albrja:
### Make new CLI group for serial workflow
- *Category*: Feature
- *JIRA issue*: https://jira.ihme.washington.edu/browse/MIC-7066

### Changes and notes
-make new CLI group for workflow
-move files, package, and tests for new group
-add new tests for CLI args


src/vivarium_cluster_tools/core/jobmon/client.py
```diff
@@ -224,3 +229,51 @@ def run_workflow(
 def count_completed_tasks(workflow: Workflow) -> int:
     """Count tasks in *workflow* whose ``final_status`` is :data:`JOBMON_STATUS_DONE`."""
     return sum(1 for t in workflow.tasks.values() if t.final_status == JOBMON_STATUS_DONE)
+
+
+def bind_and_run_workflow(
```
albrja, line 232: This got moved and is now public method. There is a very valid argument we could move jobmon_config directory out of psimulate and it would be a top level package since psimulate and this new group will use this method from it.

src/vivarium_cluster_tools/dagger/runner.py
```diff
@@ -0,0 +1,113 @@
...
+    build_workflow_from_config,
+)
+from vivarium_cluster_tools.psimulate.workflow_config.config import WorkflowConfig
+from vivarium_cluster_tools.psimulate.workflow_config.serialization import (
+    workflow_config_to_dict,
+)
+from vivarium_cluster_tools.psimulate.workflow_config.utilities import WORKFLOW_ARGS_FILENAME
+
+
+def workflow_main(
```
albrja: Moved from psimulate package.

setup.py
```diff
@@ -49,6 +49,7 @@
         "psutil",
         "layered_config_tree",
         "pytest-xdist",
+        "sphinx>=5.3,<9.0.0",
```
albrja, line 53: This is a temporary hotfix due to new version of docutils and jobmon-core package requirements.

albrja:
Code Reviewer
PR Review: Albrja/feature/mic 7066/new cli group (PR #341)
Branch confirmation

Head (feature): albrja/feature/mic-7066/new-cli-group
Base (target): albrja/feature/mic-6938/no-time-increase (not main — this stacks on the MIC-6938 branch)
Local HEAD (3488764 temporary sphinx pin) matches the PR head, diff is 9 commits / 48 files / +820/-647
Summary
The PR carves a new top-level CLI group dagger out of psimulate for multi-step Jobmon workflows. The bulk of the diff is a directory move (src/vivarium_cluster_tools/psimulate/workflow_config/* → src/vivarium_cluster_tools/dagger/workflow_config/*), plus new files dagger/cli.py, dagger/runner.py, and dagger/__init__.py. The deleted psimulate workflow subcommand is replaced by dagger run, which adds --name and --default-environment flags. notifications.py is promoted to the top-level package and gains a required command_label parameter so the Slack message can distinguish "psimulate run" from "dagger run". A new shared helper bind_and_run_workflow lives in psimulate/jobmon_config/client.py and is consumed by both runners. Tests are relocated under tests/dagger/ and a new parametrized CLI test file lands.

Design
Four parallel step-type registries can drift — STEP_TYPE_YAML_PARSERS ([parsing.py:384](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/parsing.py#L384)), STEP_TYPE_API_FNS ([builder.py:30](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/builder.py#L30)), STEP_TYPE_YAML_SERIALIZERS ([serialization.py:139](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/serialization.py#L139)), and the implicit validator binding inside each get_*_step_tasks. Adding a new step type means edits in four files with nothing to enforce co-registration. Fix: a single StepHandler dataclass bundling (parser, validator, builder, serializer) plus one STEP_HANDLERS: dict[str, StepHandler].

dagger ↔ psimulate are not peers — dagger/workflow_config/interface.py:35-41 and several other dagger files import from psimulate.{COMMANDS, branches, jobs, paths, cluster, jobmon_config}. The naming implies sibling status; the imports show strict consumption. Either rename / restructure (move the genuinely-shared bits — cluster/validation.py, cluster/interface.py, paths.py, jobmon_config/, branches.py, COMMANDS — into a vct/engine/ package both CLIs import from), or call the dependency out explicitly in dagger/__init__.py. The shared-engine move is the cleaner long-term answer; flagging now while names are still fresh.

bind_and_run_workflow violates client.py's own contract — [client.py:1-16](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/client.py#L1-L16) says it is the "single point of contact for the Jobmon SDK." The new helper at [client.py:234-279](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/client.py#L234-L279) imports get_workflow_timeout_seconds from psimulate.cluster.interface, mixing SLURM allocation logic into the Jobmon facade. Fix: make seconds_until_timeout a parameter of bind_and_run_workflow (each runner already lives in the SLURM context and can compute it), or relocate the helper to a workflow_lifecycle.py that legitimately layers above both. At minimum, update the module docstring.

WorkflowConfig.__post_init__ validates a queue by feeding validate_runtime_and_queue a placeholder runtime — [config.py:208-209](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/config.py#L208-L209) calls validate_runtime_and_queue("01:00:00", self.queue). The literal is a sentinel; a future tightening of runtime validation will silently break workflow-level validation. Fix: extract validate_queue(queue) in psimulate/cluster/validation.py and call that from both sites.

Two-phase project/queue resolution leaks throughout — ResourceConfig accepts project: str | None and queue: str | None, then _validate_common ([validation.py:33-41](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/validation.py#L33-L41)) gates them, then to_native_specification ([config.py:147-151](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/config.py#L147-L151)) re-checks with isinstance(...). Resolve fallbacks during WorkflowConfig construction so ResourceConfig.{project,queue} become non-Optional and the runtime isinstance gate disappears.

**options: Any discards Click's type information — [cli.py:85-117](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/cli.py#L85-L117). Mixes options["verbose"] (KeyError if absent) and options.get("name") (silently None); the decorator–runner contract is invisible. Click already passes options as kwargs — declare them as named parameters: def run(config_path, name, project, queue, output_directory, default_environment, max_attempts, resume, verbose, with_debugger). Aligns with [feedback_no_hidden_defaults](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/memory/feedback_no_hidden_defaults.md).

Sequential step dependencies are O(N×M) — [builder.py:78-82](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/builder.py#L78-L82). For two adjacent simulation steps of N tasks each, the dependency graph is N² edges. Correct, and likely fine for current workflows (typical post-step is single-task), but worth knowing about. Follow-up only — a no-op barrier task is the standard fix when scale bites.

notifications.py placement — Slack-specific module sitting at top level as notifications.py, with env var still PSIMULATE_SLACK_BOT_TOKEN. Either rename to slack.py (no integrations/ until a second channel materializes — YAGNI), or update the env var name if it isn't pinned by deployment infrastructure.

Maintainability
dagger/runner.py:62-63 vs psimulate/runner.py:313 — same MD5 recipe (hashlib.md5(str(root).encode()).hexdigest()[:8]), different assembled-string conventions. The "8 chars" is unexplained at both sites. Extract a small helper (see design finding 7 / DRY finding 1).

_validate_common is asymmetric about type checks — [config.py:79-84](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/config.py#L79-L84) explicitly type-checks requires_archive_node; nothing else (cores, memory_gb, hardware) gets a parallel check. Either pick a side or document why archive-node is special (YAML truthy-string parsing burned someone?).

or-short-circuit fallbacks silently swallow falsy values — [parsing.py:460-462, 502, 504](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/parsing.py#L460). name="", default_environment="", or YAML max_attempts: 0 would all fall through. CLI guards max_attempts with IntRange(min=1), but the parsing-level fallback applies to YAML too. Per [feedback_no_hidden_defaults](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/memory/feedback_no_hidden_defaults.md), replace with cli if cli is not None else workflow.get(...).

Round-trip is lossy at the default value — [serialization.py:60](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/serialization.py#L60) drops backup_freq when it equals DEFAULT_BACKUP_FREQ_SECONDS. Same for sim_verbosity != 0 ([line 62](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/serialization.py#L62)). The runner advertises configuration.yaml as reusable ([runner.py:98-99](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L98-L99)); a future change to the default silently shifts behavior for replayed configs.

notebook.cwd fallback assumes resolved path — [interface.py:431](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/interface.py#L431). path.parent of a relative path is Path(".") (the worker's cwd, not the user's). Today the parser resolves the path before reaching here, but Python-API callers (the very thing this module advertises) won't have that. Either path.resolve().parent here, or document the precondition in the docstring at [interface.py:411-413](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/interface.py#L411-L413).

send_slack_notification's new command_label is positional-eligible — [notifications.py:14-19](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/notifications.py#L14-L19). External importers calling positionally (workflow_name, status, monitoring_url, results_dir) would now silently bind monitoring_url to command_label. Make every parameter keyword-only with *, — matches how both in-repo call sites already invoke it.

dagger/runner.py:87-91 — error message reads oddly ("(expected 'D' for DONE)" presents the constant by raw value); the constant's docstring at [client.py:55-56](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/jobmon_config/client.py#L55-L56) already covers the encoding. Slim to f"Workflow finished with non-success status '{wf_status}'."

tests/dagger/test_runner.py imports inside test bodies ([test_runner.py:29-33, 77](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/dagger/test_runner.py#L29-L33)) — not the project's convention. Hoist to module top so import errors surface at collection.

DRY
Worth extracting: workflow_args recipe — hashlib.md5(str(root).encode()).hexdigest()[:8] plus {prefix}_{name}_{hash}[_{timestamp}] assembly at [dagger/runner.py:61-63](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L61-L63) and [psimulate/runner.py:313-314](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/psimulate/runner.py#L313-L314). One small helper in psimulate/jobmon_config/client.py (or the engine package from design finding 2). Also addresses design finding 7.

Worth extracting: _base_yaml prelude in serializers — all five serialize_*_step_to_yaml functions in [serialization.py:30-124](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/serialization.py#L30-L124) start with the same 8-line block (name, optional type, resources.to_dict(), optional environment). A _base_yaml(step, *, step_type) helper would shrink each by 8 lines and keep all five aligned when (e.g.) a new step-level field like description shows up.

Inconsistency, not duplication — parse_*_step_from_yaml parsers handle the environment key two ways: bash/simulation/python use kwargs["environment"] = data.get("environment") unconditionally; pytest/notebook use if "environment" in data: kwargs["environment"] = .... Pick one. The downstream consumers (builder.py:73-75, serialization) handle both shapes, but the asymmetry is gratuitous.

Skip: the validate_X / ensure_output_directory_exists / get_single_command_task tails on four interface fns ([interface.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/interface.py)), the per-type parser kwargs literals, the notebook-vs-python flag conversion (different rules), and _validate_required_paths call sites. The explicitness is the contract.

Tests
Highest-value gap: workflow_main is untested. [tests/dagger/test_runner.py](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/dagger/test_runner.py) covers only write_workflow_configuration. The brand-new runner at [dagger/runner.py:31-92](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L31-L92) has three independent branches with no unit coverage: timestamped workflow_args generation (fresh run), WORKFLOW_ARGS_FILENAME round-trip (resume), and the RuntimeError path on wf_status != JOBMON_STATUS_DONE. With client.bind_and_run_workflow and build_workflow_from_config patched, three short tests would cover all three branches.

tests/dagger/test_cli.py — no --resume test. Add a parametrize ([], False), (["--resume"], True) asserting mock_main.call_args.kwargs["resume"].

No --verbose count test. Parametrize over ([], 0), (["-v"], 1), (["-vv"], 2) and assert call_kwargs["verbose"].

coerce_to_full_path is only tested with an already-created absolute target ([test_cli.py:113-133](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/tests/dagger/test_cli.py#L113-L133)). Add a relative-path-from-monkeypatch.chdir(tmp_path) case asserting resolution to (tmp_path / "out_rel").resolve().

Empty-string and falsy CLI overrides aren't pinned — covers the same or-short-circuit issue as Maintainability finding 3. Add one test confirming the documented behavior (or pinning down rejection).

tests/dagger/test_e2e.py:293-297 is fragile — set(locations) == location_dirs would break if any non-location dir lands under output_dir (a logs/ directory, future Jobmon sentinels, etc.). Use set(locations).issubset(location_dirs) and explicit allow-list any non-location dirs.

tests/dagger/test_runner.py:69 asserts max_attempts == 2 without naming DEFAULT_MAX_ATTEMPTS. Import the constant.

tests/test_notifications.py — asserts the command_label substring but only with "dagger run"; the substring "run" is too generic to detect a regression that drops the label. Parametrize over "psimulate run" and "dagger run", and pin location in the rendered text.

Deleted tests/psimulate/test_cli.py / test_runner.py — all deleted assertions targeted the now-removed psimulate workflow subcommand and have equivalents in the new dagger tests. Nothing material lost.

Documentation
utilities.rst missing in the new Sphinx tree — docs/source/api_reference/dagger/workflow_config/index.rst globs *, so adding utilities.rst with .. automodule:: vivarium_cluster_tools.dagger.workflow_config.utilities would auto-include. As is, public symbols like WORKFLOW_ARGS_FILENAME (referenced by [runner.py:26](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L26)) are unrendered.

CHANGELOG under-describes breaking changes — [CHANGELOG.rst:1-5](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/CHANGELOG.rst#L1-L5) has one bullet. Add: (a) removal of psimulate workflow, (b) notifications.send_slack_notification move to top-level + new required command_label.

notifications.py module docstring is stale — "Slack notification support for parallel-workflow CLIs" ([notifications.py:1](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/notifications.py#L1)). Reword to mention both psimulate and dagger.

load_workflow_config docstring — [parsing.py:418-456](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/parsing.py#L418-L456) calls out the default_environment None-falls-back-to-YAML rule but not the parallel rule for max_attempts ([line 504](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/parsing.py#L504)). Mirror the note.

get_*_step_tasks "Returns" sections are uninformative — every one renders "The Jobmon tasks produced by the step." ([interface.py:154-156, 250-252, 333-335, 417-419](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/interface.py#L154)). For simulation: "One Jobmon task per (input_draw, random_seed, branch) tuple"; for the others: "A list containing a single Jobmon task wrapping the rendered command."

builder.py:71-72 "cached api_kwargs" comment — ParsedStep.api_kwargs is a dataclass field, not a cache. Reword to "...so the ParsedStep's api_kwargs is not mutated in place."

dagger/runner.py:59-61 workflow_args comment explains what but not why-timestamp-not-UUID (the value surfaces in Jobmon's UI, so human-readability matters). One-line note would help.

pipeline_automation_plan.md still references psimulate workflow in ~30+ places. If kept as historical context, add a banner; if meant to be current, rewrite.

README.rst doesn't mention dagger — the new console script in setup.py isn't surfaced. If dagger is a first-class user feature, add a short subsection.

Functionality
write_workflow_configuration is unconditional on resume — [dagger/runner.py:66](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L66) writes configuration.yaml on both fresh and resume paths. On resume with new CLI overrides, this overwrites the original record. That may be intentional (record-what-was-used-this-time) but it loses provenance of the original run. Consider either skipping the write on resume, or writing a side file (configuration.resume.yaml or appending a history block).

Resume can fail silently if WORKFLOW_ARGS_FILENAME is corrupt — [dagger/runner.py:56](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L56): workflow_args_path.read_text().strip(). If the file exists but is empty (interrupted previous run between line 73 write and the OS flush), workflow_args = "" which Jobmon will likely reject downstream with a confusing message, not "your prior run's args file is empty — start fresh." Cheap fix: validate non-empty after .strip() and produce a directed error.

Workflow-args timestamp uses second resolution — [dagger/runner.py:61](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/runner.py#L61) strftime("%Y%m%d_%H%M%S"). Two fresh dagger run invocations against the same output dir within one second produce identical workflow_args. Unlikely in production, plausible in tests. Not blocking — Jobmon would then treat the second as a resume, which is "wrong" but recoverable.

output_directory: null in YAML triggers Path(None) — [parsing.py:464](vscode-webview://006i0prfhf5et33j75a80ran1o9buk0diqansps1hslm1ov3te5s/src/vivarium_cluster_tools/dagger/workflow_config/parsing.py#L464): "output_directory" in workflow is True but workflow["output_directory"] is None. Falls into Path(None).resolve() → TypeError. Either check workflow.get("output_directory") is not None, or wrap with a clearer error.

dagger/workflow_config/parsing.py:262 — [str(Path(p).resolve()) for p in raw_path] will TypeError for an int element (YAML path: [1, 2]). Path-list elements should be string-validated in validate_pytest_step.

Minor Nits
dagger/cli.py:8 — Sphinx .. click:: directive style matches the existing psimulate/cli.py directive; no change needed.
dagger/workflow_config/parsing.py:178-182 — simulation example sets backup_freq: 1800 (exactly the default); either drop it or note that it's optional.
dagger/workflow_config/interface.py:441-448 — notebook param isinstance(value, bool) ordering is correct (bool before int) but invisible glue; consider a literal mapping table to make it explicit.
Overall
Solid PR. The package extraction itself is mostly mechanical and uncontroversial. The most material areas to address before merge:

Test coverage for workflow_main (Tests #1) — net-new code with three untested branches.
Step-handler unification (Design #1) — the four parallel registries are the single biggest source of future drift.
bind_and_run_workflow placement / coupling (Design #3) — client.py's contract is broken by the new helper.
**options → named parameters in dagger run (Design #6) — tightens the click-decorator-to-runner contract.
CHANGELOG entries for the two breaking changes (Documentation #2).
The naming question (dagger vs psimulate as peers when one strictly consumes the other — Design #2) is worth discussing now while it's cheap, even if you defer the actual engine-package move.

albrja:
Key questions to be answered by reviewers:
-what should the new CLI group be? Dagger was a fun placeholder name that I do like but something like workflow would be just fine but I personally find it too generic
-do we like "core" for containing the jobmon and cluster packages that both this new group and psimulate will consume?

src/vivarium_cluster_tools/core/jobmon/client.py
rmudambi: I think that likely makes sense. In principle, it seems that dagger shouldn't need to call anything from psimulate except the part that creates the Jobmon tasklist for simulation steps. Let's talk through the implications though.

src/vivarium_cluster_tools/dagger/workflow_config/__init__.py
rmudambi, line 1: Maybe this package could just be called `config` rather than `workflow_config`?

src/vivarium_cluster_tools/dagger/config/config.py
```diff
@@ -19,14 +19,21 @@
...
 
-REQUIRED_WORKFLOW_FIELDS = {"name", "steps"}
+REQUIRED_WORKFLOW_FIELDS = {"steps"}
+"""Top-level workflow fields that must appear in the YAML file.
+
+``name``, ``project``, ``queue``, and ``output_directory`` are each
+required overall but may be provided via either the YAML or a CLI
+override; their presence is validated by
+:func:`~vivarium_cluster_tools.dagger.workflow_config.parsing.load_workflow_config`
+rather than by :meth:`WorkflowConfig.parse_yaml_file`."""
```
rmudambi, line 32:
```suggestion
"""Top-level workflow fields that must appear in the YAML file rather than a CLI override."""
```

src/vivarium_cluster_tools/dagger/workflow_config/parsing.py
```diff
@@ -418,38 +418,46 @@ def parse_step_from_yaml(
...
+        CLI override for the workflow name.
     project
         CLI override for the project field.
     queue
         CLI override for the queue field.
     output_directory
         CLI override for the output directory.
+    default_environment
+        CLI override for the default_environment field. ``None`` from the
+        CLI falls back to the YAML value (which itself may be absent).
```
rmudambi:
```suggestion
        CLI override for the default_environment field.
```

src/vivarium_cluster_tools/dagger/runner.py
```diff
@@ -0,0 +1,115 @@
...
+from vivarium_cluster_tools.core.notifications import send_slack_notification
+from vivarium_cluster_tools.dagger.workflow_config.builder import build_workflow_from_config
+from vivarium_cluster_tools.dagger.workflow_config.config import WorkflowConfig
+from vivarium_cluster_tools.dagger.workflow_config.serialization import (
+    workflow_config_to_dict,
+)
+from vivarium_cluster_tools.dagger.workflow_config.utilities import WORKFLOW_ARGS_FILENAME
+
+
+def workflow_main(
```
rmudambi: Maybe a better name would be `run()` or `run_workflow()`?

src/vivarium_cluster_tools/dagger/runner.py
```diff
@@ -0,0 +1,115 @@
...
+
+    if wf_status != client.JOBMON_STATUS_DONE:
+        raise RuntimeError(
+            f"Workflow finished with status '{wf_status}' "
+            f"(expected '{client.JOBMON_STATUS_DONE}' for DONE)."
+        )
+    logger.info(f"Workflow completed successfully. Results in {output_root}")
+
+
+def write_workflow_configuration(output_root: Path, workflow_config: WorkflowConfig) -> None:
```
rmudambi: Is this referenced anywhere else? If not it could be inlined in the method above. At a minimum, it should be made private.

rmudambi (approved):

stevebachmeier (approved):

